PR #2, opened 2018-09-19
XRSessionMode proposal

toji:
Finding this one tough to keep focused on JUST introducing the `XRSessionMode` concept, but that's probably necessary for not getting lost in the weeds, Gonna add some comments inline re: thing I'm thinking about that I'd like opinions on.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -61,7 +61,7 @@ The basic steps most WebXR applications will go through are:
 
 The UA will identify an available physical unit of XR hardware that can present imagery to the user, referred to here as an "XR device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard/Daydream or Samsung Gear VR). It may also represent devices without stereo-presentation capabilities but with more advanced tracking, such as ARCore/ARKit-compatible devices. Any queries for XR capabilities or functionality are implicitly made against this device.
 
-> **Non-normative Note:** If there are multiple XR devices available, the UA will need to pick which one to expose. The UA is allowed to use any criteria it wishes to select which device is used, including settings UI that allows users to manage device priority. Calling `navigator.xr.supportsSession` or `navigator.xr.requestSession` with `{ immersive: false }` should **not** trigger device-selection UI, however, as this would cause many sites to display XR-specific dialogs early in the document lifecycle without user activation.
+> **Non-normative Note:** If there are multiple XR devices available, the UA will need to pick which one to expose. The UA is allowed to use any criteria it wishes to select which device is used, including settings UI that allows users to manage device priority. Calling `navigator.xr.supportsSession` or `navigator.xr.requestSession` with `{ mode: 'inline' }` should **not** trigger device-selection UI, however, as this would cause many sites to display XR-specific dialogs early in the document lifecycle without user activation.
 
 It's possible that even if no XR device is available initially, one may become available while the application is running, or that a previously available device becomes unavailable. This will be most common with PC peripherals that can be connected or disconnected at any time. Pages can listen to the `devicechange` event emitted on `navigator.xr` to respond to changes in device availability after the page loads. (XR devices already available when the page loads will not cause a `devicechange` event to be fired.) `devicechange` fires an event of type `Event`.
 
@@ -75,21 +75,23 @@ The first thing that any XR-enabled page will want to do is query to determine i
 
 Testing to see if the device supports the capabilities the application needs is done via the `navigator.xr.supportsSession` call, which takes a dictionary describing the desired functionality and returns a promise which resolves if the device can support those properties and rejects otherwise. Querying for support this way is necessary because it allows the application to detect what XR features are available without actually engaging the sensors or beginning presentation, which can incur significant power or performance overhead on some systems and may have side effects such as launching a status tray, launching a storefront, or terminating another application's access to XR hardware. Calling `navigator.xr.supportsSession` should also not interfere with any running XR applications on the system.
 
-There are two primary classes of XR content that can be displayed to the user:
+There are three primary classes of XR content that can be displayed to the user:
 
-**Immersive**: Indicated with the `immersive: true` dictionary argument. Immersive content is presented directly to the XR device (for example: displayed on a VR headset). Immersive presentation must be started within a user activation event or within another callback that has been explicitly indicated to allow immersive presentation. As a result, if immersive content is supported, the application will usually want to add some UI to trigger activation of "XR Presentation Mode", where the application can begin sending imagery to the device. 
+**Immersive VR**: Indicated with the `mode: 'immersive-vr'` dictionary argument. Immersive VR content is presented directly to the XR device (for example: displayed on a VR headset). Immersive presentation must be started within a user activation event or within another callback that has been explicitly indicated to allow immersive presentation. As a result, if immersive VR content is supported, the application will usually want to add some UI to trigger activation of "XR Presentation Mode", where the application can begin sending imagery to the device.
 
-**Non-Immersive (in-page)**: The default mode, but can be explicitly requested with the `immersive: false` dictionary argument. Non-immersive content does not have the ability to display on the XR device, but is able to access device tracking information and use it to render content on the page. This technique, where a scene rendered to the page is responsive to device movement, is sometimes referred to as "Magic Window" mode. It's especially useful for mobile devices, where moving the device can be used to look around a scene. Devices like Tango phones and tablets with 6DoF tracking capabilities may expose them via non-immersive sessions even if the hardware is not capable of immersive, stereo presentation.
+**Immersive AR**: Indicated with the `mode: 'immersive-ar'` dictionary argument. Immersive AR content function largely like Immersive VR content, with the primary difference being that it guarantees that the users environment will be visible and aligned with the rendered content. Examples include HoloLens or Magic Leap-style headsets, as well as phone-based AR systems like ARCore and ARKit. (It should also be noted that immersive VR content may make the environment visible anyway, especially on transparent displays. See [Handling non-opaque displays](#handling-non-opaque-displays) for more details.) Additionally in the future access to environmental data (such as hit testing) will be permitted. Like Immersive VR content, Immersive AR content must be started within a user activation event or within another callback that has been explicitly indicated to allow immersive presentation.
 
-In the following examples we will focus on using immersive content, and cover non-immersive content use in the [`Advanced Functionality`](#non-immersive-sessions-magic-windows) section. With that in mind, this code checks for supports of immersive content, since we want the ability to display imagery on a device like a headset.
+**Inline**: The default mode, but can be explicitly requested with the `mode: 'inline'` dictionary argument. inline content does not have the ability to display on the XR device, but is able to access device tracking information and use it to render content on the page. This technique, where a scene rendered to the page is responsive to device movement, is sometimes referred to as "Magic Window" mode. It's especially useful for mobile devices, where moving the device can be used to look around a scene. Devices like Tango phones and tablets with 6DoF tracking capabilities may expose them via inline sessions even if the hardware is not capable of immersive, stereo presentation.
+
+In the following examples we will focus on using immersive VR content, and cover inline and immersive AR content use in the [`Advanced Functionality`](#inline-sessions) section. With that in mind, this code checks for supports of immersive VR content, since we want the ability to display imagery on a device like a headset.
 
 ```js
 async function checkForXRSupport() {
-  // Check to see if there is an XR device available that's capable of immersive
+  // Check to see if there is an XR device available that supports immersive VR
   // presentation (for example: displaying in a headset). If the device has that
   // capability the page will want to add an "XR" button to the page (similar to
-  // a "Fullscreen" button) that starts the display of immersive content.
-  navigator.xr.supportsSession({ immersive: true }).then(() => {
+  // a "Fullscreen" button) that starts the display of immersive VR content.
+  navigator.xr.supportsSession({ mode: 'immersive-vr' }).then(() => {
     var enterXrBtn = document.createElement("button");
     enterXrBtn.innerHTML = "Enter VR";
     enterXrBtn.addEventListener("click", beginXRSession);
@@ -110,7 +112,7 @@ Clicking the button in the previous sample will attempt to acquire an `XRSession
 function beginXRSession() {
   // requestSession must be called within a user gesture event
   // like click or touch when requesting an immersive session.
-  navigator.xr.requestSession({ immersive: true })
+  navigator.xr.requestSession({mode: 'immersive-vr' })
       .then(onSessionStarted)
       .catch(err => {
         // May fail for a variety of reasons. Probably just want to
@@ -120,7 +122,7 @@ function beginXRSession() {
 }
 ```
 
-Only one immersive session per XR hardware device is allowed at a time across the entire UA. Any non-immersive sessions are suspended when an immersive session is active. Non-immersive sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it.
+Only one immersive session per XR hardware device is allowed at a time across the entire UA. Any inline sessions are suspended when an immersive session is active. Inline sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it.
 
 Once the session has started, some setup must be done to prepare for rendering.
 - A `XRFrameOfReference` should be created to establish a coordinate system in which `XRDevicePose` data will be defined. See the [Spatial Tracking Explainer](spatial-tracking-explainer.md) for more information.
@@ -243,7 +245,7 @@ function drawScene(view, pose) {
 
 The UA may temporarily "suspend" a session at any time. While suspended a session has restricted or throttled access to the XR device state and may process frames slowly or not at all. Suspended sessions can be reasonably be expected to be resumed at some point, usually when the user has finished performing whatever action triggered the suspension in the first place.
 
-The UA may suspend a session if allowing the page to continue reading the headset position represents a security or privacy risk (like when the user is entering a password or URL with a virtual keyboard, in which case the head motion may infer the user's input), or if other content is obscuring the page's output. Additionally, non-immersive sessions are suspended while an immersive session is active.
+The UA may suspend a session if allowing the page to continue reading the headset position represents a security or privacy risk (like when the user is entering a password or URL with a virtual keyboard, in which case the head motion may infer the user's input), or if other content is obscuring the page's output. Additionally, inline sessions are suspended while an immersive session is active.
 
 While suspended the page may either refresh the XR device at a slower rate or not at all, and poses queried from the device may be less accurate. If the user is wearing a headset the UA is expected to present a tracked environment (a scene which remains responsive to user's head motion) when the page is being throttled to prevent user discomfort.
 
@@ -291,7 +293,7 @@ function onSessionEnd() {
 }
 ```
 
-The UA may end a session at any time for a variety of reasons. For example: The user may forcibly end presentation via a gesture to the UA, other native applications may take exclusive access of the XR hardware device, or the XR hardware device may become disconnected from the system. Additionally, if the system's underlying XR device changes (signaled by the `devicechange` event on the `navigator.xr` object) any active `XRSession`s will be ended. This applies to both Immersive and Non-immersive sessions. Well behaved applications should monitor the `end` event on the `XRSession` to detect when the UA forces the session to end.
+The UA may end a session at any time for a variety of reasons. For example: The user may forcibly end presentation via a gesture to the UA, other native applications may take exclusive access of the XR hardware device, or the XR hardware device may become disconnected from the system. Additionally, if the system's underlying XR device changes (signaled by the `devicechange` event on the `navigator.xr` object) any active `XRSession`s will be ended. This applies to both Immersive and Inline sessions. Well behaved applications should monitor the `end` event on the `XRSession` to detect when the UA forces the session to end.
 
 ```js
 xrSession.addEventListener('end', onSessionEnd);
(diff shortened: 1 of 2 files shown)


explainer.md
```diff
@@ -95,17 +95,19 @@ navigator.xr.addEventListener('devicechange', checkForXR);
...
+**Inline**: The default mode, but can be explicitly requested with the `mode: 'inline'` dictionary argument. Inline sessions do not have the ability to display immersive content on the `XRDevice` but are able to access device tracking information and use it to render content on the page. This technique, where a scene rendered to the page is responsive to device movement, is sometimes referred to as "Magic Window" mode. It's especially useful for mobile devices, where moving the device can be used to look around a scene. Devices like Tango phones and tablets with 6DoF tracking capabilities may expose them via inline sessions even if the hardware is not capable of immersive, stereo presentation. Inline sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it.
```
toji: There's a few things I want to do here, eventually. One is scrub this document of the term "magic window" and focus entirely on advancing understanding of the "inline" concept. The other is restructure the entire doc so that inline and AR mode aren't presented as footnotes in the "advanced" section. I think that shift may be more appropriate when we tackle the Universal Rendering Path concept, though.

explainer.md
```diff
@@ -95,17 +95,19 @@ navigator.xr.addEventListener('devicechange', checkForXR);
...
+**Immersive**: Requested with the `mode: 'immersive'` dictionary argument. Immersive sessions present content directly to the XR hardware device (frequently a headset), enabling immersive presentation. Immersive sessions are considered "exclusive". Only one exclusive session per XR hardware device is allowed at a time across the entire UA, and any non-exclusive sessions are suspended while an exclusive session is active. Immersive sessions must be created within a user activation event or within another callback that has been explicitly indicated to allow exclusive session creation.
```
toji: Sticking with the term "immersive" for the moment to avoid conflating this with the ongoing bikeshed, but I'm not particularly attached to the term. It is a bit weird to have to explain that the immersive modes are exclusive in how they access the hardware, so perhaps reverting to that term simplifies the message.

explainer.md
```diff
@@ -95,17 +95,19 @@ navigator.xr.addEventListener('devicechange', checkForXR);
...
+
+**Immersive AR**: Requested with the `mode: 'immersive-ar'` dictionary argument. These sessions function largely like standard Immersive sessions, with the primary difference being that they guarantee that the users environment will be visible and aligned with the rendered content. (It should also be noted that standard immersive sessions may make the environment visible anyway, especially on transparent displays. See [Handling non-opaque displays](#handling-non-opaque-displays) for more details.) Additionally in the future access to environmental data (such as hit testing) will be permitted. Like standard Immersive sessions, Immersive AR sessions are considered "exclusive" and must be created within a user activation event or within another callback that has been explicitly indicated to allow exclusive session creation.
```
toji: Would *love* a sanity check on if this description of AR mode (and the additional details below) actually click. I feel like I'm being overly simplistic.

Also when we start talking about "Immersive AR" I find myself backed into talking about "standard immersive sessions", which.... ewe.

explainer.md
```diff
@@ -95,17 +95,19 @@ navigator.xr.addEventListener('devicechange', checkForXR);
...
 
-In the following examples we will focus on using immersive sessions, and cover non-immersive session use in the [`Advanced Functionality`](#non-immersive-sessions-magic-windows) section. With that in mind, we ask here if the `XRDevice` supports immersive sessions, since we want the ability to display imagery on the headset.
+In the following examples we will focus on using immersive sessions, and cover inline session use in the [`Advanced Functionality`](#inline-sessions-magic-windows) section. With that in mind, we ask here if the `XRDevice` supports immersive sessions, since we want the ability to display imagery on the headset.
```
toji: LOTS of straightforward "non-immersive"->"inline" search and replace here. Apologies for the noise.

explainer.md
```diff
@@ -346,50 +348,69 @@ function beginXRSession() {
...
+```js
+function checkARSupport() {
+  // Check to see if the UA can support an AR sessions.
+  return xrDevice.supportsSession({ mode: 'immersive-ar' })
+      .then(() => { console.log("AR content is supported!"); })
+      .catch((reason) => { console.log("AR content is not supported: " + reason); });
+}
+```
+
+In addition to displaying on a dedicated XR hardware device like a standard Immersive session, the UA may choose instead to display an AR session output on a 2D screen, such as with [ARKit](https://developer.apple.com/arkit/) and [ARCore](https://developers.google.com/ar/) compatible devices. When displayed in that way, the UA must transition to a mode where the AR session's output is shown exclusively upon creation, hiding the rest of the page. (Similar to the transition that happens when invoking the `requestFullscreen` API.) The UA must also provide a way of exiting that mode and returning to the normal view of the page, at which point the AR session must end.
```
toji: A bunch of questions this paragraph brings up for me.

Do we need to communicate to the developer that they're going to be presenting to a 2D screen rather than a headset? If so should that be before or after session creation? Is that pit-of-success-y? Until we add DOM overlays (which I'm studiously avoiding talking about here) it largely feels like the main difference between headset AR and phone AR from a developer's perspective is the available input devices.

And FWIW when it comes to DOM overlays and such I still think there's possibly room for a 4th mode that's explicitly targeted at Phone AR-style interaction, but if we went that route I'd say that it would fit comfortably as a 4th mode in the enum ('immersive-ar-with-overlay' or something silly like that) and that this mode could still function exactly as described.

explainer.md
```diff
@@ -1008,14 +1029,20 @@ partial interface Navigator {
...
 dictionary XRSessionCreationOptions {
-  boolean immersive = false;
+  XRSessionMode mode = "inline";
   XRPresentationContext outputContext;
 };
 
 [SecureContext, Exposed=Window] interface XRSession : EventTarget {
   readonly attribute XRDevice device;
-  readonly attribute boolean immersive;
+  readonly attribute XRSessionMode mode;
```
toji: We both agreed previously that "mode" makes the most sense when the sessions are mutable, which is still something I plan to propose but am avoiding here. Sticking with the "mode" terminology anyway to potentially reduce churn.

explainer.md
NellWaliczek: I'd leave it out any references to the overlay or differentiating handheld devices for now.  Let's keep this change scoped, but make sure we've got an issue filed to cover it.

explainer.md
```diff
@@ -346,50 +348,69 @@ function beginXRSession() {
...
-  // Request a non-immersive session for magic window rendering.
+  // Request an inline session for magic window rendering.
   xrDevice.requestSession({ outputContext: magicWindowCtx })
       .then(OnSessionStarted)
       .catch((reason) => { console.log("requestSession failed: " + reason); });
 }
 ```
 
-The UA may reject requests for a non-immersive sessions for a variety of reasons, such as the inability of the underlying hardware to provide tracking data without actively rendering to the device. Pages should be designed to robustly handle the inability to acquire non-immersive sessions. `XRDevice.supportsSession()` can be used if a page wants to test for non-immersive session support before attempting to create the `XRSession`.
+The UA may reject requests for an inline sessions for a variety of reasons, such as the inability of the underlying hardware to provide tracking data without actively rendering to the device. Pages should be designed to robustly handle the inability to acquire inline sessions. `XRDevice.supportsSession()` can be used if a page wants to test for inline session support before attempting to create the `XRSession`.
```
NellWaliczek: Do we want to require inline sessions to be supported?  Or do we need to resolve the poseless sessions thing first?

explainer.md
```diff
@@ -432,7 +453,7 @@ The `targetRay` will never be `null`. It's value will differ based on the type o
...
   * `'tracked-pointer'` indicates that the target ray originates from either a handheld device or other hand-tracking mechanism and represents that the user is using their hands or the held device for pointing. The exact orientation of the ray relative to a given device should follow platform-specific guidelines if there are any. In the absence of platform-specific guidance or a physical device, the target ray should most likely point in the same direction as the user's index finger if it was outstretched.
-  * `'screen'` indicates that the input source was an interaction with the canvas element associated with a non-immersive session's output context, such as a mouse click or touch event. See [Magic Window Input](#magic_window_input) for more details.
+  * `'screen'` indicates that the input source was an interaction with the canvas element associated with a inline session's output context or an immersive AR session, such as a mouse click or touch event. See [Magic Window Input](#magic_window_input) for more details.
```
NellWaliczek: Technically, you could build a VR session on ARCore....

explainer.md
NellWaliczek: I wonder if that'll end up causing a bikeshedding rathole... no strong opinion right now

explainer.md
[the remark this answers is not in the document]
toji: Yeah, that's one of those "definitely intend to change this but trying to keep this PR focused" things. (I'm finding a LOT of those as I go)

explainer.md
toji: I wondered about that. Felt to me like we needed to resolve poseless first.

explainer.md
toji: > Actually, it totally doesn't have to be and that's key. There's absolutely no reason an AR Core device would need to be in Daydream mode. Just sayin'

I hear what you're saying, but I can't help but wonder about the developer intent behind it. If I put a button on your page that looks like a headset, but when you click it all you get is a fancy fullscreen mode with non-obvious 6DoF tracking it's likely not what you or the developer wanted? (It's frequently hard to differentiate between orientation tracking and true 6DoF when you're sitting in place and viewing a life-sized scene through a handheld screen.)

Plus, if that's the behavior you wanted it seems like you could just ask for immersive-ar and draw over the full screen? (Handwaving about lost performance.)

explainer.md
toji: Even if we're avoiding the mutable sessions path for the time being now, I've come around to "mode" being the more natural thing to type in many cases, so I'm sticking to it.

toji:
Rebased and updated. I **know** I'm gonna get some people up in arms with the `immersive-vr` and `immersive-ar` enum suggestions, but honestly the more I think about it the more I think we really just need names that suggest to the developer what they WANT to do, rather than how the system sees it. (This mentality inspired by naming Frames of Reference things like 'stationary', because that's how you expect the user to experience your content.)

Would appreciate another look, and if there's nothing terrifying I'll push the PR to the `webxr` repo to watch people tear each other apart over the mode names. (Don't intend to merge prior to the FoR changes landing if I can help it.)

explainer.md
```diff
@@ -410,7 +431,7 @@ The `targetRay` will never be `null`. It's value will differ based on the type o
...
   * `'tracked-pointer'` indicates that the target ray originates from either a handheld device or other hand-tracking mechanism and represents that the user is using their hands or the held device for pointing. The exact orientation of the ray relative to a given device should follow platform-specific guidelines if there are any. In the absence of platform-specific guidance or a physical device, the target ray should most likely point in the same direction as the user's index finger if it was outstretched.
-  * `'screen'` indicates that the input source was an interaction with the canvas element associated with a non-immersive session's output context, such as a mouse click or touch event. See [Magic Window Input](#magic_window_input) for more details.
+  * `'screen'` indicates that the input source was an interaction with the canvas element associated with a inline session's output context or an immersive AR session, such as a mouse click or touch event. See [Inline Input](#inline_input) for more details.
```
NellWaliczek: This is only true if the immersive AR session isn't in an headset, yeah?

NellWaliczek: A general comment... do we need to provide some indication of the session not being headworn for UI purposes?  Or is that being punted to the DOM overlay issue?

explainer.md
```diff
@@ -75,21 +75,23 @@ The first thing that any XR-enabled page will want to do is query to determine i
...
+**Inline**: The default mode, but can be explicitly requested with the `mode: 'inline'` dictionary argument. inline content does not have the ability to display on the XR device, but is able to access device tracking information and use it to render content on the page. This technique, where a scene rendered to the page is responsive to device movement, is sometimes referred to as "Magic Window" mode. It's especially useful for mobile devices, where moving the device can be used to look around a scene. Devices like Tango phones and tablets with 6DoF tracking capabilities may expose them via inline sessions even if the hardware is not capable of immersive, stereo presentation.
+
+In the following examples we will focus on using immersive VR content, and cover inline and immersive AR content use in the [`Advanced Functionality`](#inline-sessions) section. With that in mind, this code checks for supports of immersive VR content, since we want the ability to display imagery on a device like a headset.
```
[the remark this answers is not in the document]
toji: That was my thoughts on it. (Re: waiting for mutable output contexts) General concern is that I want to introduce the simplest mode with the least amount of setup first, which happens to be immersive VR. That way the document doesn't start off with the already confusing concept of output contexts. When we have mutable output contexts (and especially if we make WebGL contexts a valid output context) that progression will hopefully become less problematic.

explainer.md
```diff
@@ -120,7 +122,7 @@ function beginXRSession() {
 }
 ```
 
-Only one immersive session per XR hardware device is allowed at a time across the entire UA. Any non-immersive sessions are suspended when an immersive session is active. Non-immersive sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it.
+Only one immersive session per XR hardware device is allowed at a time across the entire UA. Any inline sessions are suspended when an immersive session is active. Inline sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it.
```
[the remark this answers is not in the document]
toji: Added a note early in the document to explain this.

explainer.md
toji: That specific spec issue feels like it's better handled in the context of the DOM overlay issue, yes.

toji:
Updated again, being a bit more aggressive about reworking how certain concepts are presented now while still steering clear of a full document restructuring. I'm hoping to defer moving the inline sections around until we have the mutable outputContext changes in place.

explainer.md
```diff
@@ -100,17 +108,17 @@ async function checkForXRSupport() {
...
 
-Clicking the button in the previous sample will attempt to acquire an `XRSession` by calling `navigator.xr.requestSession()` method. This returns a promise that resolves to an `XRSession` upon success. When requesting a session, the capabilities that the returned session must have are passed in via a dictionary, exactly like the `supportsSession` call. If `supportsSession` resolved for a given dictionary, then calling `requestSession` with the same dictionary values should be reasonably expected to succeed, barring external factors (such as `requestSession` not being called in a user activation event for an immersive session.) The UA is ultimately responsible for determining if it can honor the request.
+Clicking the button in the previous sample will attempt to acquire an `XRSession` by calling `navigator.xr.requestSession()` method. This returns a promise that resolves to an `XRSession` upon success. When requesting a session, the capabilities that the returned session must have, including it's XR mode, are passed in via a dictionary.
```
NellWaliczek: nit: this block and the paragraph above seem to be a bit duplicative.

explainer.md
```diff
@@ -120,7 +128,9 @@ function beginXRSession() {
 }
 ```
 
-Only one immersive session per XR hardware device is allowed at a time across the entire UA. Any non-immersive sessions are suspended when an immersive session is active. Non-immersive sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it.
+If `supportsSessionMode` resolved for a given mode, then requesting a session with the same mode should be reasonably expected to succeed, barring external factors (such as `requestSession` not being called in a user activation event for an immersive session.) The UA is ultimately responsible for determining if it can honor the request.
+
+Only one immersive session per XR hardware device is allowed at a time across the entire UA. Any inline sessions are suspended when an immersive session is active. Inline sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it.
```
NellWaliczek: nit: Any inline -> All inline

explainer.md
```diff
@@ -246,7 +256,7 @@ function drawScene(view, pose) {
 
 The UA may temporarily "suspend" a session at any time. While suspended a session has restricted or throttled access to the XR device state and may process frames slowly or not at all. Suspended sessions can be reasonably be expected to be resumed at some point, usually when the user has finished performing whatever action triggered the suspension in the first place.
 
-The UA may suspend a session if allowing the page to continue reading the headset position represents a security or privacy risk (like when the user is entering a password or URL with a virtual keyboard, in which case the head motion may infer the user's input), or if other content is obscuring the page's output. Additionally, non-immersive sessions are suspended while an immersive session is active.
+The UA may suspend a session if allowing the page to continue reading the headset position represents a security or privacy risk (like when the user is entering a password or URL with a virtual keyboard, in which case the head motion may infer the user's input), or if other content is obscuring the page's output. Additionally, inline sessions are suspended while an immersive session is active.
```
NellWaliczek: >Additionally, inline sessions are suspended while an immersive session is active.

A bit weird that we're saying this in two places, but ok :)

explainer.md
```diff
@@ -294,17 +304,45 @@ function onSessionEnd() {
...
+The UA may end a session at any time for a variety of reasons. For example: The user may forcibly end presentation via a gesture to the UA, other native applications may take exclusive access of the XR hardware device, or the XR hardware device may become disconnected from the system. Additionally, if the system's underlying XR device changes (signaled by the `devicechange` event on the `navigator.xr` object) any active `XRSession`s will be ended. This applies to both Immersive and Inline sessions. Well behaved applications should monitor the `end` event on the `XRSession` to detect when the UA forces the session to end.
```
NellWaliczek: >This applies to both Immersive and Inline sessions.

I wish I remember why we decided this... It seems like this might be a pit of failure....  as in, developers won't monitor the event and the page will be very badly behaved if a headset is plugged in after the fact.

explainer.md
```diff
@@ -294,17 +304,45 @@ function onSessionEnd() {
...
+```js
+function beginXRSession() {
+  // requestSession must be called within a user gesture event
+  // like click or touch when requesting an immersive session.
+  navigator.xr.requestSession({mode: 'immersive-ar'})
+      .then(onSessionStarted);
+}
+```
+
+This provides a session that behaves much like the typical Immersive VR sessions described above with a few key behavioral differences. The primary distinction between an "immersive-vr" and "immersive-ar" session is that the latter guarantees that the user's environment is visible and that rendered content will be aligned to the environment. The exact nature of the visibility is hardware-dependent, and communicated by the `XRSession`'s `environmentBlendMode` attribute. AR sessions will never report an `environmentBlendMode` of `opaque`. See [Handling non-opaque displays](#handling-non-opaque-displays) for more details.
```
NellWaliczek: nit: the typical Immersive VR -> the Immersive VR

explainer.md
```diff
@@ -294,17 +304,45 @@ function onSessionEnd() {
...
+```js
+function checkARSupport() {
+  // Check to see if the UA can support an AR sessions.
+  return navigator.xr.supportsSessionMode('immersive-ar')
+      .then(() => { console.log("AR content is supported!"); })
+      .catch((reason) => { console.log("AR content is not supported: " + reason); });
+}
+```
+
+The UA may choose to present the immersive AR session's content via any type of display, including dedicated XR hardware (for devices like HoloLens or Magic Leap) or 2D screens (for APIs like [ARKit](https://developer.apple.com/arkit/) and [ARCore](https://developers.google.com/ar/)). In all cases the session takes exclusive control of the display, hiding the rest of the page if necessary. On a phone screen, for example, this would mean that the session's content should be displayed in a mode that is distinct from standard page viewing, similar to the transition that happens when invoking the `requestFullscreen` API. The UA must also provide a way of exiting that mode and returning to the normal view of the page, at which point the immersive AR session must end.
```
NellWaliczek: I really like how  you reworded this paragraph 👍

explainer.md
```diff
@@ -324,59 +362,59 @@ function beginXRSession() {
...
+Immersive and inline sessions can use the same render loop, but there are some differences in behavior to be aware of. The sessions may run their render loops at at different rates. During immersive sessions the UA runs the rendering loop at the XR device's native refresh rate. During inline sessions the UA runs the rendering loop at the refresh rate of page (aligned with `window.requestAnimationFrame`.) The method of computation of `XRView` projection and view matrices also differs between immersive and inline sessions, with inline sessions taking into account the output canvas dimensions and possibly the position of the users head in relation to the canvas if that can be determined.
```
NellWaliczek: Oh geez.. this just made me realize another issue with the Diorama frame of reference... the refresh rate may be wrong O_O

explainer.md
```diff
@@ -324,59 +362,59 @@ function beginXRSession() {
...
+function beginInlineXRSession() {
+  // Request an inline session in order to render to the page.
+  navigator.xr.requestSession({ outputContext: inlineCtx })
       .then(OnSessionStarted)
       .catch((reason) => { console.log("requestSession failed: " + reason); });
 }
 ```
 
-The UA may reject requests for a non-immersive sessions for a variety of reasons, such as the inability of the underlying hardware to provide tracking data without actively rendering to the device. Pages should be designed to robustly handle the inability to acquire non-immersive sessions. `navigator.xr.supportsSession()` can be used if a page wants to test for non-immersive session support before attempting to create the `XRSession`.
+The UA should not reject requests for an inline session unless the page's feature policy prevents it. `navigator.xr.supportsSessionMode()` can still be used if a page wants to test if inline session are allowed.
```
NellWaliczek: Interesting... this seems like a new change... can you give me a bit of background on this change?

explainer.md
```diff
@@ -410,7 +448,7 @@ The `targetRay` will never be `null`. It's value will differ based on the type o
...
   * `'tracked-pointer'` indicates that the target ray originates from either a handheld device or other hand-tracking mechanism and represents that the user is using their hands or the held device for pointing. The exact orientation of the ray relative to a given device should follow platform-specific guidelines if there are any. In the absence of platform-specific guidance or a physical device, the target ray should most likely point in the same direction as the user's index finger if it was outstretched.
-  * `'screen'` indicates that the input source was an interaction with the canvas element associated with a non-immersive session's output context, such as a mouse click or touch event. See [Magic Window Input](#magic_window_input) for more details.
+  * `'screen'` indicates that the input source was an interaction with a session's output context canvas element, such as a mouse click or touch event. Only applicable for inline sessions or an immersive AR session being displayed on a 2D screen. See [Screen Input](#screen_input) for more details.
```
NellWaliczek: Maybe add "instead of from within a headset" to make it really clear?

explainer.md
```diff
@@ -649,9 +687,9 @@ function onUpdateScene() {
 
 The above sample is optimized for dragging items in the scene around using input sources that have a gripMatrix. It would also be possible to add further script logic to use the target ray properties to position items in the world - this is left as an exercise for the reader.
 
-### Magic Window Input
+### Screen Input
 
-When using a non-immersive session, pointer events on the canvas that created the `outputContext` passed during the session request are monitored. `XRInputSource`s are generated in response to allow unified input handling with immersive mode controller or gaze input.
+When using an inline session or an immersive AR session on a 2D screen, pointer events on the canvas that created the session's `outputContext` are monitored. `XRInputSource`s are generated in response to allow unified input handling with immersive mode controller or gaze input.
```
NellWaliczek: same as above

explainer.md
toji: Are you referring specifically to the feature policy bit? I felt that was fairly natural conclusion given the way feature policies work. It would be weird, to me, if the parent page of an iframe said "No XR for you!" and the page was still happily able to create `XRSessions`, regardless of functionality level.

explainer.md
toji: Augh! But what if the "2D screen" is a floating quad that you're viewing in your headset?!? WHY ARE THESE THINGS SO HARD TO DESCRIBE ACCURATELY?????

toji:
Moved to main repo and merged. Closing.

PR #4, opened 2018-09-29
Remove XRDevice

toji:
Straightforward pull request for removing a formal `XRDevice` object from the explainer entirely. Sorry I'm getting it out so late. If you get a chance to take a look let me know what you think.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -21,10 +21,10 @@ That said, given the shared subject matter many of the same concepts are represe
 ### Goals
 Enable XR applications on the web by allowing pages to do the following:
 
-* Detect available VR/AR devices.
-* Query the devices capabilities.
-* Poll the device’s position and orientation.
-* Display imagery on the device at the appropriate frame rate.
+* Detect if XR capabilities are available.
+* Query the XR devices capabilities.
+* Poll the XR device’s pose (position and orientation).
+* Display imagery on the XR device at the appropriate frame rate.
 
 ### Non-goals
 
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -21,10 +21,10 @@ That said, given the shared subject matter many of the same concepts are represe
 ### Goals
 Enable XR applications on the web by allowing pages to do the following:
 
-* Detect available VR/AR devices.
-* Query the devices capabilities.
-* Poll the device’s position and orientation.
-* Display imagery on the device at the appropriate frame rate.
+* Detect if XR capabilities are available.
+* Query the XR devices capabilities.
```
NellWaliczek: Should this now be "Request an XR device with specific capabilities"?

explainer.md
```diff
@@ -21,10 +21,10 @@ That said, given the shared subject matter many of the same concepts are represe
...
+* Poll the XR device’s pose (position and orientation).
```
NellWaliczek: Do we want to make this statement a bit more generic?  In the sense that it's not just pose, but also input, etc.

explainer.md
```diff
@@ -50,75 +50,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
  1. Use the session to run a render loop that produces graphical frames to be displayed on the XR device.
  1. Continue producing frames until the user indicates that they wish to exit XR mode.
  1. End the XR session.
 
-### Acquiring a Device
+### XR hardware
 
-The first thing that any XR-enabled page will want to do is request an `XRDevice` and, if one is available, advertise XR functionality to the user. (For example, by adding a button to the page that the user can click to start XR content.)
+The UA will identify an available physical unit of XR hardware that can present imagery to the user, referred to here as an "XR device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard/Daydream or Samsung Gear VR). It may also represent devices without stereo-presentation capabilities but with more advanced tracking, such as ARCore/ARKit-compatible devices. Any queries for XR capabilities or functionality are implicitly made against this device.
```
NellWaliczek: Do we want to call out the diorama case?

explainer.md
```diff
@@ -50,75 +50,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
-      console.error('Requesting XR device failed:', err);
-    }
-  });
-}
-```
-
-Future revisions of the API may add filter criteria to `navigator.xr.requestDevice`.
-
-> **Non-normative Note:** If there are multiple XR devices available, the UA will need to pick which one to return. The UA is allowed to use any criteria it wishes to select which device is returned, including settings UI that allows users to manage device priority. Calling `navigator.xr.requestDevice` should not trigger device-selection UI, however, as this would cause many sites to display XR-specific dialogs early in the document lifecycle without user activation.
+> **Non-normative Note:** If there are multiple XR devices available, the UA will need to pick which one to expose. The UA is allowed to use any criteria it wishes to select which device is used, including settings UI that allows users to manage device priority. Calling `navigator.xr.supportsSession` or `navigator.xr.requestSession` should **not** trigger device-selection UI, however, as this would cause many sites to display XR-specific dialogs early in the document lifecycle without user activation.
```
NellWaliczek: >Calling `navigator.xr.supportsSession` or `navigator.xr.requestSession` should **not** trigger device-selection UI, however, as this would cause many sites to display XR-specific dialogs early in the document lifecycle without user activation.

I wish we could make this normative. *sigh*

explainer.md
```diff
@@ -50,75 +50,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
 
-**Immersive**: Requested with the `immersive: true` dictionary argument. Immersive sessions present content directly to the `XRDevice`, enabling immersive presentation. Only one immersive session per XR hardware device is allowed at a time across the entire UA. Immersive sessions must be created within a user activation event or within another callback that has been explicitly indicated to allow immersive session creation.
+Testing to see if the device supports the capabilities the application needs is done via the `navigator.xr.supportsSession` call, which takes a dictionary describing the desired functionality and returns a promise which resolves if the device can support those properties and rejects otherwise. Querying for support this way is necessary because it allows the application to detect what XR features are available without actually engaging the sensors or beginning presentation, which can incur significant power or performance overhead on some systems and may have side effects such as launching a status tray or storefront.
```
NellWaliczek: Can we add another consequence regarding forcing other applications currently using the XR hardware to exit?

explainer.md
```diff
@@ -50,75 +50,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+There are two primary classes of XR content that can be displayed to the user:
 
-### Detecting and advertising XR mode
+**Immersive**: Indicated with the `immersive: true` dictionary argument. Immersive content is presented directly to the XR device (for example: displayed on a VR headset). Immersive presentation must be started within a user activation event or within another callback that has been explicitly indicated to allow immersive presentation. As a result, if immersive content is supported, the application will usually want to add some UI to trigger activation of "XR Presentation Mode", where the application can begin sending imagery to the device.
```
NellWaliczek: For whatever reason, this is showing up as a diff.  You may want your PR description to call out very explicitly that you aren't doubling down on any particular language and and that there are more PRs coming that will address those issues separately.

explainer.md
```diff
@@ -50,75 +50,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+**Non-Immersive (in-page)**: The default mode, but can be explicitly requested with the `immersive: false` dictionary argument. Non-immersive content does not have the ability to display on the XR device, but is able to access device tracking information and use it to render content on the page. This technique, where a scene rendered to the page is responsive to device movement, is sometimes referred to as "Magic Window" mode. It's especially useful for mobile devices, where moving the device can be used to look around a scene. Devices like Tango phones and tablets with 6DoF tracking capabilities may expose them via non-immersive sessions even if the hardware is not capable of immersive, stereo presentation.
```
NellWaliczek: Yeaaaah, double for this section ^_^;;;;

explainer.md
```diff
@@ -50,75 +50,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
 
-In the following examples we will focus on using immersive sessions, and cover non-immersive session use in the [`Advanced Functionality`](#non-immersive-sessions-magic-windows) section. With that in mind, we ask here if the `XRDevice` supports immersive sessions, since we want the ability to display imagery on the headset.
+In the following examples we will focus on using immersive content, and cover non-immersive content use in the [`Advanced Functionality`](#non-immersive-sessions-magic-windows) section. With that in mind, this code checks for supports of immersive content, since we want the ability to display imagery on a device like a headset.
```
NellWaliczek: Q: are you planning on leaving "inline" as part of the advanced section?  

Personally, i'm still leaning pretty strongly towards pulling input into its own explainer.  And I've been thinking the same might  be good for legitimately "Advanced topics" like the rendering optimization sections.  That said, `inline` doesn't feel like it falls into that bucket, tbh...

explainer.md
```diff
@@ -50,75 +50,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
 
 ```js
-let xrDevice = null;
-
-async function onXRAvailable(device) {
-  xrDevice = device;
-
-  // Many XRDevices are capable of immersive presentation to
+async function checkForXRSupport() {
+  // Many XR devices are capable of immersive presentation to
```
NellWaliczek: This sentence feels awkward... especially in the context of desktop browsers...?

explainer.md
```diff
@@ -129,15 +101,17 @@ async function onXRAvailable(device) {
...
+Clicking the button in the previous sample will attempt to acquire an `XRSession` by callling `navigator.xr.requestSession()` method. This returns a promise that resolves to an `XRSession` upon success. When requesting a session, the capabilities that the returned session must have are passed in via a dictionary, exactly like the `supportsSession` call. If `supportsSession` resolved for a given dictionary, then calling `requestSession` with the same dictionary values should be reasonably expected to succeed, barring external factors (such as `requestSession` not being called in a user activation event for an immersive session.) The UA is ultimately responsible for determining if it can honor the request.
```
NellWaliczek: Another one of those paragraphs that's showing up as having a bigger diff than it really is and which might be confusing to other folks.

explainer.md
```diff
@@ -146,6 +120,9 @@ function beginXRSession() {
       });
 }
 ```
+
+Only one immersive session per XR hardware device is allowed at a time across the entire UA. Any non-immersive sessions are suspended when an immersive session is active. Non-immersive sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it.
```
NellWaliczek: Q: Should we file an issue to cover the explorations we're doing regarding mutability?

explainer.md
```diff
@@ -178,35 +155,35 @@ function onSessionStarted(session) {
...
+**XR Centric:** The app's primary use case is displaying XR content, and as such it doesn't mind initializing resources in an XR-centric fashion, which may include asking users to select a headset as soon as the app starts. An example would be a game which is dependent on XR presentation and input. These types of applications can avoid the need to call `makeXRCompatible` and the possible context loss that it may trigger by passing the XR device that the context will be used with as a context creation argument.
 
 ```js
-let gl = glCanvas.getContext("webgl", { compatibleXRDevice: xrDevice });
+let gl = glCanvas.getContext("webgl", { xrCompatible: true });
 ```
 
-Ensuring context compatibility with an `XRDevice` through either method may have side effects on other graphics resources in the page, such as causing the entire user agent to switch from rendering using an integrated GPU to a discrete GPU.
+Ensuring context compatibility with an XR device through either method may have side effects on other graphics resources in the page, such as causing the entire user agent to switch from rendering using an integrated GPU to a discrete GPU.
```
NellWaliczek: As you pointed out, disconnecting XR hardware may make a webgl context no longer compatible.  We should call out the expected behavior.  I was assuming it wouldn't cause an adaptor change, but that the `xrCompatible` property would no longer be true?
Related to that, I think that means we need to require `makeXRCompatible` to be run again even if the same device is plugged back into the same adaptor.
And now that `XRDevice` is gone... what does that mean for an `XRSession` that was using it? (incidentally, the answer to this question may end up being the biggest nail in the coffin for mutable sessions...)

explainer.md
toji: I see your sigh and raise you a **_siiiigghhhh_**

explainer.md
toji: Yeah, I read it again and agree. Feels like something that was once reasonable and had been tweaked ever so slightly over the course of multiple changes into something that was just kinda weird.

explainer.md
toji: I don't actually think that's accurate?

explainer.md
toji: Was planning on that for a follow-up PR.

toji:
Updated the PR, but a fair amount of the feedback will simply need to be part of the comment when I push the PR to the main repo.

toji:
Migrated to https://github.com/immersive-web/webxr/pull/405

PR #5, opened 2018-10-04
Mutable Sessions proposal

toji:
I don't intend to push `mutable-session-explainer.md` to the webxr repo, I'm only committing it here to get feedback on the design and iterate quickly. Once it looks like we're feeling good about the direction I'll change this into an actually diff against the main explainer doc that will eventually be pushed to the webxr repo.


mutable-session-explainer.md
```diff
@@ -0,0 +1,75 @@
...
+// I demand bikeshedding by combat! Last one alive decides the enum names.
+enum XRImmersiveMode {
+  "none",
+  "virtual",
+  "augmented"
+};
+
+[SecureContext, Exposed=Window] partial interface XR {
+  // supportsSession has been removed.
+  Promise<XRSession> requestSession(XRImmersiveMode mode);
```
NellWaliczek: For whatever reason, I had to read the whole document before I realized that this was the value for the "target" immersive mode; the one that would be entered upon calling `requestImmersiveMode()`.  I'm absolutely not bikeshedding.. (ok a little) but could we rename this to something like `targetMode` or gosh I don't know what... but something about how it's set up made me thing that was the mode it was going to start in when created.

toji:
Definitely not planning on throwing this out until we the tracking systems PR has landed.

Haven't thought much about how this interacts with navigation, TBH. I doubt it really changes the landscape there much, though.

toji:
The borked table at the end of the doc, till I figure out what's wrong:

|FrameOfRef                      |Inline|Virtual|Augmented|
|--------------------------------|------|-------|---------|
|**Stationary/Eye-level**        |Always|Always |Always   |
|**Stationary/Floor-level**      |Always|Always |Always(?)|
|**Stationary/Position-disabled**|Always|Always |Always(?)|
|**Diorama (theory)**            |Always|Never  |Never    |
|**Bounded**                     |Never |H/W dep|H/W dep  |
|**Unbounded**                   |Never |H/W dep|Always(?)|

toji:
Suspending work on this PR for a while given the complications with FoR/Layer sharing between modes that we've identified. Might get back to it later if there's demand for it. In the meantime a couple of ideas from here (such as mutable outputContexts) will likely get broken out into their own issues/PR.

PR #8, opened 2018-10-12
Mutable outputContext proposal

### LINKED ISSUE #6: PR Needed: Mutable outputContext
Write a PR to change the outputContext to a read/write attribute on the `XRSession` rather than a readonly attribute that is set during session creation. Needs to describe what happens if an inline session has no output context and how to resolve an outputContext being assigned to multiple sessions. PR Should close the issue about outputContext lifetimes.

toji:
Follow up to #2, makes outputContext Mutable. Fixes #6

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -130,7 +130,7 @@ In this sample, the `beginXRSession` function, which is assumed to be run by cli
 
 If `supportsSessionMode` resolved for a given mode, then requesting a session with the same mode should be reasonably expected to succeed, barring external factors (such as `requestSession` not being called in a user activation event for an immersive session.) The UA is ultimately responsible for determining if it can honor the request.
 
-Only one immersive session per XR hardware device is allowed at a time across the entire UA. All inline sessions are suspended when an immersive session is active. Inline sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it.
+Only one immersive session per XR hardware device is allowed at a time across the entire UA. All inline sessions are [suspended](#handling-suspended-sessions) when an immersive session is active. Inline sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it.
 
 Once the session has started, some setup must be done to prepare for rendering.
 - An `XRReferenceSpace` should be created to establish a space in which `XRViewerPose` data will be defined. See the [Spatial Tracking Explainer](spatial-tracking-explainer.md) for more information.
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -342,28 +342,31 @@ The UA may choose to present the immersive AR session's content via any type of
...
 
-A `XRPresentationContext` can only be supplied imagery by an `XRSession`, though the exact behavior depends on the scenario in which it's being used.
+An `XRPresentationContext` can only be supplied imagery by an `XRSession`, though the exact behavior depends on the scenario in which it's being used. The context is associated with a session by setting the `XRSession`'s `outputContext` attribute to the desired `XRPresentationContext` object. An `XRPresentationContext` cannot be used with multiple `XRSession`s simultaneously, so when an `XRPresentationContext` is set as the `outputContext` for a session, any session it was previously set as the `outputContext` for will have it's `outputContext` set to `null`.
```
NellWaliczek: Does this mean we need an event?

explainer.md
```diff
@@ -342,28 +342,31 @@ The UA may choose to present the immersive AR session's content via any type of
...
+In order to mirror WebXR content to the page, the session's `outputContext` must be set to a `XRPresentationContext`. Once the session has a valid `outputContext` any content displayed on the headset will then be mirrored into the canvas associated with the `outputContext`.
 
 When mirroring only one eye's content will be shown, and it should be shown without any distortion to correct for headset optics. The UA may choose to crop the image shown, display it at a lower resolution than originally rendered, and the mirror may be multiple frames behind the image shown in the headset. The mirror may include or exclude elements added by the underlying XR system (such as visualizations of room boundaries) at the UA's discretion. Pages should not rely on a particular timing or presentation of mirrored content, it's really just for the benefit of bystanders or demo operators.
```
NellWaliczek: I know this section is about mirroring, but it seems related to how we think about the diorama problem.  How would we describe that behavior?

explainer.md
```diff
@@ -380,9 +383,11 @@ These scenarios can make use of inline sessions to render tracked content to the
...
 
-Similar to mirroring, to make use of this mode an `XRPresentationContext` is provided as the `outputContext` at session creation time with an inline session. At that point content rendered to the `XRSession`'s `baseLayer` will be rendered to the canvas associated with the `outputContext`. The UA is also allowed to composite in additional content if desired. In the future, if multiple `XRLayers` are used their composited result will be what is displayed in the `outputContext`. Requests to create an inline session without an output context will be rejected.
+Similar to mirroring, to make use of this mode the session's `outputContext` must be set. At that point content rendered to the `XRSession`'s `baseLayer` will be rendered to the canvas associated with the `outputContext`. The UA is also allowed to composite in additional content if desired. In the future, if multiple `XRLayers` are used their composited result will be what is displayed in the `outputContext`.
```
NellWaliczek: Yeah, this isn't quite how we'll want to describe it for diorama support....

NellWaliczek:
Are you waiting on anything from me for this one?

toji:
I haven't been actively pursuing this one recently in favor of landing some other items first. Worth looping back on. May even tie into the state object discussed on #13? Dunno.

toji:
Updated this to sit on top of the `XRRenderState` PR, now that that one has moved to the webxr repo.

Looking back through the previous comments, I'm really not sure what the diorama concerns are about? Sorry. Assuming that we can come to an understanding on that, though, I'd like to move this over to the WebXR repo as well.

explainer.md
NellWaliczek: Specifically, dioramas are likely to be both inline and sterescopic, which means it's not just a straightforward matter of rendering into the canvas's default framebuffer.

toji: Ah, I see. So a couple of thoughts on that: For one, I don't see mirroring ever working out in a diorama mode, since you're actually seeing from a different person's point of view? Maybe stereo mirroring could work, but it would still be a bit wonky.

As for inline stereo, yes, you're exactly right. It's not just a matter of rendering into the default backbuffer. That's why the rendering path for inline is still to render into the `XRWebGLLayer`'s framebuffer (which is NOT the default canvas backbuffer except maybe in certain polyfill situations, in which case that fact will be hidden), and then the page compositor picks that up and uses it as the content for the output canvas instead. In the case of stereo inline content we would render two views to the layer's framebuffer just like with a headset, and then let the page compositor do whatever is needed to translate those two viewports to the display properly. (zSpace has already successfully proof-of-concepted this by blitting the layer framebuffer's viewports into a Quadbuffer stereo texture for their hardware.)

This hasn't been stated too explicitly in the explainer thus far because it wasn't thought of as an immediate need, but the fact that we're having this conversation highlights to me that it should be called out.

NellWaliczek: Ah yeah, I agree that mirroring is a non-issue for diorama because it's inline.  I was more concerned about the spec language about the very nature of inline stereo, but it sounds like you understand my concern.  Looking forward to see that in the revision. :)

NellWaliczek:
Have you thought more about how to remove the need for a secondary canvas? I'd kinda like to have a sense of how those two things relate before we merge this, though I'm not 100% tied to that.

explainer.md
NellWaliczek: Poke?

toji:
Post F2F this has been migrated to webxr/504

PR #9, opened 2018-11-01
More generic use of Coordinate Systems

toji:
A few different things have gotten my thinking about how to make the concept of "tracking stuff" in WebXR more generic/extensible. Some examples:

- Verbiage weirdness when talking about querying viewers/views
- The conversation about the body locked frame of reference and how it felt like an unfortunate one-off
- Reviewing some input concepts
- Thinking about Anchors
- Thinking about Hit testing reticles
- Name bikeshedding
- Not wanting to have the number of ways we exposed tracked objects explode (because it IS a core part of the API)
- How we integrate better with various native libraries

So the idea that came out of all of that that I wanted to feel out is simply pushing `XRCoordinateSystem` to a more visible, central place in the API (which probably further justifies the spatial systems bikeshedding, FWIW). Basically making it so that ANYTHING that is tracked spatially is represented as a coordinate system and all of them use a single function to get it's spatial state. (Querying the views array is still a special case because it needs to be.) Some highlights:

 - XRFrame is two functions: `getViewer()` for the views array and `getPose()` for all coordinate systems.
 - Input devices are represented as two coordinate systems: target and grip.
 - The grip coordinate system being nullable nicely allows us to communicate "I don't have a grip pose" vs "I don't have a grip pose _right now_"
 - You can construct an `XRRay` from any `XRPose`. Useful for things like forcing gaze cursors in all cases.

This WIP patch is more focused on IDL and sample code changes than explainer text changes at the moment, but I don't want to invest a lot of time into the text until I've sanity checked it with you. It builds off of a prior but not yet merged PR I've got up on the main repo because the changes it introduced nicely support this change.

### DIFF
--- a/input-explainer.md
+++ b/input-explainer.md
@@ -25,15 +25,17 @@ The properties of an XRInputSource object are immutable. If a device can be mani
 
 ### Input poses
 
-Each input source can query a `XRInputPose` using the `getInputPose()` function of any `XRFrame`. Getting the pose requires passing in the `XRInputSource` you want the pose for, as well as the `XRReferenceSpace` the pose values should be given in, just like `getViewerPose()`. `getInputPose()` may return `null` in cases where tracking has been lost (similar to `getViewerPose()`), or the given `XRInputSource` instance is no longer connected or available.
+Each input source provides two `XRSpace`s, which can be used to query an `XRPose` using the `getPose()` function of any `XRFrame`. Getting the pose requires passing in the `XRSpace` you want the pose for, as well as the `XRSpace` the returned pose should be relative to (which may be an `XRReferenceSpace`). Just like `getViewerPose()`, `getPose()` may return `null` in cases where tracking has been lost, or the `XRSpace`'s `XRInputSource` instance is no longer connected or available.
 
-The `gripTransform` is an `XRRigidTransform` into a space where if the user was holding a straight rod in their hand it would be aligned with the negative Z axis (forward) and the origin rests at their palm. This enables developers to properly render a virtual object held in the user's hand. For example, a sword would be positioned so that the blade points directly down the negative Z axis and the center of the handle is at the origin.
+The `gripSpace` represents a space where if the user was holding a straight rod in their hand it would be aligned with the negative Z axis (forward) and the origin rests at their palm. This enables developers to properly render a virtual object held in the user's hand. For example, a sword would be positioned so that the blade points directly down the negative Z axis and the center of the handle is at the origin.
 
-If the input source has only 3DOF, the `gripTransform` may represent only a translation or rotation based on tracking capability. An example of this case is for physical hands on some AR devices which only have a tracked position. The `gripTransform` will be `null` if the input source isn't trackable. 
+If the input source has only 3DOF, the grip pose may represent only a translation or rotation based on tracking capability. An example of this case is for physical hands on some AR devices which only have a tracked position. The `gripSpace` will be `null` if the input source isn't trackable. 
 
-An input source will also provide its preferred pointing ray, given by the `XRInputPose`'s `targetRay`. The ray, which is an `XRRay` object, includes both an `origin` and `direction`, both given as `DOMPointReadOnly`s. The `origin` represents a 3D coordinate in space with a `w` component that must be 1, and the `direction` represents a normalized 3D directional vector with a `w` component that must be 0. A matrix can also be queried from the `XRRay` with the `matrix` attribute which represents the transform from a ray originating at `[0, 0, 0]` and extending down the negative Z axis to the ray described by the `XRRay`'s `origin` and `direction`. This is useful for positioning graphical representations of the ray.
+An input source will also provide its preferred pointing ray, given by the `XRInputSource`'s `targetRaySpace`. This can be used to get the input's target ray `XRPose`.
 
-The `targetRay` will never be `null`. It's value will differ based on the type of input source that produces it, which is represented by the `targetRayMode` attribute:
+An `XRRay` can be constructed from the `XRPose`'s `transform`. Rays includes both an `origin` and `direction`, both given as `DOMPointReadOnly`s. The `origin` represents a 3D coordinate in space with a `w` component that must be 1, and the `direction` represents a normalized 3D directional vector with a `w` component that must be 0. The `XRRay`'s `matrix` represents the transform from a ray originating at `[0, 0, 0]` and extending down the negative Z axis to the ray described by the `XRRay`'s `origin` and `direction`. This is useful for positioning graphical representations of the ray.
+
+The `targetCoordinateSystem` will never be `null`. It's value will differ based on the type of input source that produces it, which is represented by the `targetMode` attribute:
 
   * `'gaze'` indicates the target ray will originate at the user's head and follow the direction they are looking (this is commonly referred to as a "gaze input" device). While it may be possible for these devices to be tracked (and have a `gripTransform`), the head gaze is used for targeting. Example devices: 0DOF clicker, regular gamepad, voice command, tracked hands.
   * `'tracked-pointer'` indicates that the target ray originates from either a handheld device or other hand-tracking mechanism and represents that the user is using their hands or the held device for pointing. The exact orientation of the ray relative to a given device should follow platform-specific guidelines if there are any. In the absence of platform-specific guidance or a physical device, the target ray should most likely point in the same direction as the user's index finger if it was outstretched.
@@ -42,16 +44,16 @@ The `targetRay` will never be `null`. It's value will differ based on the type o
 ```js
 // Loop over every input source and get their pose for the current frame.
 for (let inputSource of xrInputSources) {
-  let inputPose = xrFrame.getInputPose(inputSource, xrReferenceSpace);
+  let targetRayPose = xrFrame.getPose(inputSource.targetRaySpace, xrReferenceSpace);
 
   // Check to see if the pose is valid
-  if (inputPose) {
+  if (targetRayPose) {
     // Render a visualization of the input source (see next section).
-    renderInputSource(session, inputSource, inputPose);
+    let gripPose = xrFrame.getPose(inputSource.gripCoordinateSystem, xrFrameOfRef);
+    renderInputSource(session, inputSource, gripPose);
 
     // Highlight any objects that the target ray intersects with.
-    let ray = inputPose.targetRay;
-    let hoveredObject = scene.getObjectIntersectingRay(ray.origin, ray.direction);
+    let hoveredObject = scene.getObjectIntersectingRay(new XRRay(targetRayPose.transform));
     if (hoveredObject) {
       // Render a visualization of the object that is highlighted (see below).
       drawHighlightFrom(hoveredObject, inputSource);
@@ -72,7 +74,7 @@ function onSessionStarted(session) {
     lastInputSource = event.inputSource;
   });
   session.addEventListener("inputsourceschange", ev => {
-    // Choose an appropriate default from available inputSources, such as prioritizing based on the value of targetRayMode:
+    // Choose an appropriate default from available inputSources, such as prioritizing based on the value of targetMode:
     // 'screen' over 'tracked-pointer' over 'gaze'.
     lastInputSource = computePreferredInputSource(session.getInputSources());
   });
@@ -92,10 +94,10 @@ function drawHighlightFrom(hoveredObject, inputSource) {
 function drawScene() {
   // Display only a single cursor or ray, on the most recently used input source.
   if (lastInputSource) {
-    let inputPose = xrFrame.getInputPose(lastInputSource, xrReferenceSpace);
-    if (inputPose) {
+    let targetRayPose = xrFrame.getPose(lastInputSource.targetRaySpace, xrReferenceSpace);
+    if (targetRayPose) {
       // Render a visualization of the target ray/cursor of the active input source. (see next section)
-      renderCursor(lastInputSource, inputPose)
+      renderCursor(lastInputSource, targetRayPose)
     }
   }
 }
@@ -134,11 +136,10 @@ function onSessionStarted(session) {
 }
 
 function onSelect(event) {
-  let inputPose = event.frame.getInputPose(event.inputSource, xrReferenceSpace);
-  if (inputPose) {
+  let targetRayPose = event.frame.getPose(event.inputSource.targetRaySpace, xrReferenceSpace);
+  if (targetRayPose) {
     // Ray cast into scene to determine if anything was hit.
-    let ray = inputPose.targetRay;
-    let selectedObject = scene.getObjectIntersectingRay(ray.origin, ray.direction);
+    let selectedObject = scene.getObjectIntersectingRay(new XRRay(targetRayPose.transform));
     if (selectedObject) {
       selectedObject.onSelect();
     }
@@ -146,7 +147,7 @@ function onSelect(event) {
 }
 ```
 
-Some input sources (such as those with a `targetRayMode` of `screen`) will be only be added to the list of input sources whilst a primary action is occurring. In these cases, the `inputsourceschange` event will fire just prior to the `selectstart` event, then again when the input source is removed after the `selectend` event.
+Some input sources (such as those with a `targetMode` of `screen`) will be only be added to the list of input sources whilst a primary action is occurring. In these cases, the `inputsourceschange` event will fire just prior to the `selectstart` event, then again when the input source is removed after the `selectend` event.
 
 `selectstart` and `selectend` can be useful for handling dragging, painting, or other continuous motions.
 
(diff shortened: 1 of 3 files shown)


ddorwin:
>* Input devices are represented as two coordinate systems: target and grip.

Is it weird to have separate coordinate systems for the same object (input device)? Note that https://github.com/immersive-web/webxr/blob/master/spatial-tracking-explainer.md#unbounded-frame-of-reference currently says, "Each instance of an XRCoordinateSystem does not have a fixed relationship with any other." We could change that text, but it does seem reasonable.

toji:
Whoops! I don't recall closing this... still want to talk about it a bit more

NellWaliczek:
Honestly, with the changes you've made for the transform offset PR, I'm even more unsure this PR is necessary...

toji:
I feel there's still a good potential change here that goes beyond the teleport change, though I agree that one kind of helps the transition in to the `getPose` concept presented here.

I've rebased this and cleaned it up a bit with the assumption that if it did happen it would be built atop #11 , so it's worth taking another look to see if you feel the idea is something we should pursue.

toji:
Inviting @thetuvix to take a look at this because it ties into some of what he's referring to in the comments on webxr PR 440

toji:
Updated this PR based on the discussion in the editors call and to propose a fix (or at least a pattern for fixing) webxr issue 445.

input-explainer.md
```diff
@@ -332,15 +332,14 @@ enum XRTargetRayMode {
...
   readonly attribute XRTargetRayMode targetRayMode;
 };
 
 [SecureContext, Exposed=Window]
-interface XRInputPose {
-  readonly attribute boolean emulatedPosition;
-  readonly attribute XRRay targetRay;
+interface XRInputPose : XRPose {
+  readonly attribute XRRay targetRay; // Ray version of the pose transform?
   readonly attribute XRRigidTransform? gripTransform;
```
NellWaliczek: Ok, what if we rename `gripTransform` to something like `trackedInputSourceOrigin`?  Yeah, it's long and wordy...

NellWaliczek: Actually, the more I think about it, the more I wonder for which input elements are the ones that need acceleration and velocity... is it the grip or is it the target ray?  Or both?  The answer to that will impact what we choose to do here... my inclination is to pull Alex in and ask him about this one.  IIRC he's spent a lot of time thinking about how one throws a ball in VR :)

spatial-tracking-explainer.md
```diff
@@ -369,24 +369,27 @@ partial interface XRSession {
...
+  // TODO: Room for things like velocity and acceleration in the future
 };
 
 //
 // Space
 //
 
 [SecureContext, Exposed=Window] interface XRSpace : EventTarget {
-  XRRigidTransform? getTransformTo(XRSpace other);
+  // TODO: Anything meaningful we can put here?
```
NellWaliczek: If not, we should switch this to be a Union right now.

thetuvix:
In the original immersive-web/webxr#435 PR, @toji integrated the first half of "Idea 1" from immersive-web/webxr#418, which moved to an "is-a" pattern for reference spaces and `XRSpace`, as they are fundamentally just well-known spaces:

> ### Idea 1:
> * `XRSpace` (replaces `XRCoordinateSystem`)
>   * `XRReferenceSpace` (replaces `XRFrameOfReference`)
>     * `XRStationaryReferenceSpace`
>       * `XRStationaryReferenceSpaceSubtype`
>         * `"position-disabled"`
>         * `"eye-level"`
>         * `"floor-level"`
>     * `XRBoundedReferenceSpace`
>     * `XRUnboundedReferenceSpace`

The second half of "Idea 1" had then explored what this design implies for other locatable entities like planes, faces, meshes, viewers and input sources.  Pose relationships are just one aspect of these entities, and so it may make more sense to keep a "has-a" relationship to one or more `XRSpace`s:

> Has one or more `XRSpace` attributes:
> * `XRAnchor`
>   * `XRPlaneAnchor` (with `extents` attribute, etc.)
>   * `XRFaceAnchor` (with `getGeometry()` method, etc.)
>   * `XRMeshAnchor` (with `getVertexBuffer()` method, etc.)
> * `XRViewer`
> * `XRInputSource` (has `grip` and `pointer` spaces, etc.)
> 
> This approach explicitly establishes an `XRReferenceSpace` as a special core type of `XRSpace` that you make first, within which you reason about any other space. This feels meaningful even if types like `XRInputSource` are not subtypes of `XRSpace` - reference spaces are fundamentally about core spatial reasoning, while an `XRInputSource` will have lots of other members that are not about spatial reasoning.

`XRInputSource` especially benefits from staying "has-a" because it has two meaningful spaces: the `grip` space and the `targetRay` space.

To @nellwaliczek's question about motivation for further changes here, we'd previously discussed this sort of general mechanism for durable dynamic `XRSpace`s in immersive-web/webxr#384, identifying three API features that would be naturally enabled with such support.  If we decide not to pursue that path, we should first ensure we've sketched out an alternate solution for each of those features.

### Strawman
Some proposed changes to manifest what @toji discussed in the description for input:
* Add two input source attributes: **`XRInputSource.gripSpace`** and **`XRInputSource.targetRaySpace`**.
* Add an **`XRPose.ray`** attribute.
* Remove the separate ~~`XRFrame.getInputPose`~~ function and ~~`XRInputPose`~~ interface.

Developers then call the general `XRFrame.getPose` function defined in this PR to get input poses.

**Before:**
```js
// Loop over every input source and get their pose for the current frame.
for (let inputSource of xrInputSources) {
  let inputPose = xrFrame.getInputPose(inputSource, xrReferenceSpace);

  // Check to see if the pose is valid
  if (inputPose) {
    // Render a visualization of the input source (see next section).
    renderInputSource(session, inputSource, inputPose);

    // Highlight any objects that the target ray intersects with.
    let ray = inputPose.targetRay;
    let hoveredObject = scene.getObjectIntersectingRay(ray.origin, ray.direction);
    if (hoveredObject) {
      // Render a visualization of the object that is highlighted (see below).
      drawHighlightFrom(hoveredObject, inputSource);
    }
  }
}
```

**After:**
```js
// Loop over every input source and get their pose for the current frame.
for (let inputSource of xrInputSources) {
  let gripPose = xrFrame.getPose(inputSource.gripSpace, xrReferenceSpace);
  // Check to see if the grip pose is valid
  if (gripPose) {
    // Render visualizations of the input source (see next section).
    renderInputSource(session, inputSource, gripPose);
  }

  let targetRayPose = xrFrame.getPose(inputSource.targetRaySpace, xrReferenceSpace);
  // Check to see if the target ray pose is valid
  if (targetRayPose) {
    // Highlight any objects that the target ray intersects with.
    let ray = targetRayPose.ray;
    let hoveredObject = scene.getObjectIntersectingRay(ray.origin, ray.direction);
    if (hoveredObject) {
      // Render a visualization of the object that is highlighted (see below).
      drawHighlightFrom(hoveredObject, inputSource);
    }
  }
}
```

Some benefits of this change:
* The nullability of the grip pose and target ray pose are cleanly separated at the start, eliminating a further null check on `gripTransform` within `renderInputSource`.
* Either of these spaces can be passed to a future space-relative `XRSession.requestHitTest` function, as explored in immersive-web/hit-test#31 to create a stable reticle.
* This generic `getPose` pattern then extends to getting poses for future plane/face/mesh anchor spaces, especially for highly dynamic anchors like faces for which we then won't need further `.get...Pose` functions on `XRFrame`.
* Either of these spaces can be passed to a future `XRLayer.setBasisCoordinateSystem` function to allow reprojecting a "controller layer" independently from the primary world layer.

input-explainer.md
thetuvix: There are two common scenarios for "throwing":
* Throwing an object you're logically gripping in the palm of your hand (e.g. origin of the grip space)
* Flinging an object that is leashed to the tip of your controller (e.g. origin of the target ray space)

The velocity you want to impart to the thrown object will be the velocity of the relevant space.  However, if you're flicking your wrist when you release the object, the velocity of your palm will be quite different than the velocity of the tip of your controller.  Unlike the simple translation that gets you from grip space to target ray space, non-trivial math is needed to get from a linear and angular velocity in one space to the velocities in another space, given an arbitrary center of rotation.

If we expect that apps will make direct use of this velocity data, it's likely worth just delivering the correct velocity of the app's chosen space.  It seems reasonable to just deliver those velocities as nullable values next to the position/orientation on any `XRPose`.

thetuvix: PS. Also, all professionals are supposed to establish their "brand" now.  Apparently, this is now mine:
> Here lies @thetuvix
> He really knew how to throw a ball

NellWaliczek: Made. My. Morning. 🤣

toji: Interested to hear from Alex on that point as well. It's not something that particularly concerned me, but when both @ddorwin and @NellWaliczek brought up the same concern about it independently I took it as a sign and backed away from the two spaces model. :) I'm personally comfortable with it either way.

spatial-tracking-explainer.md
toji: If we went with the two-space-input model that @thetuvix was talking about then having this as a separate type, even if it has no members, is still useful. There's precedence for that in the WebGL spec as well, where most of the "object types" are effectively opaque handles.

input-explainer.md
thetuvix: Whether we choose a one-space or n-space model for controllers, the current statement in the spec that @ddorwin quoted is already a bit too strong:

> Each instance of an XRCoordinateSystem does not have a fixed relationship with any other.

For example, in an outside-in system such as Lighthouse tracking, a free-space `XRAnchor` that the app created would not adjust after creation, and so it would have a "fixed relationship" to the `XRBoundedReferenceSpace`.  We are not promising that all spaces are independent, just that apps should not assume any two particular spaces have a fixed relationship unless they have some other reason to know that to be the case.  Agreed that we should consider representing such fixed relationships structurally within the API where possible, balanced against any simplicity that results from directly getting a durable space for meaningful dynamic points in the world (e.g. an input source's target ray).

Now that we have the ability to create spaces with an `originOffset`, perhaps there's a middle ground where apps needing a `targetRaySpace` (e.g. for durable reticles as in immersive-web/hit-test#31) would just get the input source's root `XRSpace` and then create an offset target ray space themselves?  That could still enable the various scenarios from immersive-web/webxr#384, though I'm not sure yet if that results in less net complexity.

In general, if something like the target ray is so important that we'd have done the math for you in `XRInputPose` to deliver it in your target space, perhaps that's a sign that it's also valuable enough to track directly over time as its own space?  The answer there may emerge as we get to the bottom of the hit-test and anchors design.

One side precedent for treating poses and rays similarly is how Unity treats its [`Transform` class](https://docs.unity3d.com/ScriptReference/Transform.html).  This class has `position` and `rotation` members, but it also has `forward`/`up`/`right` members.  This lets code either:
* Reason about `camera.transform.position` and `camera.transform.rotation` to place some object at the camera, or
* Reason about `camera.transform.position` and `camera.transform.forward` to raycast from the camera, without requiring extra math.

There are two options then for giving apps a ray for a given `XRPose`:
* A `ray` property on `XRPose` returning today's `XRRay`: `pose.ray.origin` and `pose.ray.direction`
* Just flatten `forward`/`up`/`right` members onto `XRPose`: `pose.position` and `pose.forward`

toji:
Nell and I discussed this again today and I think we both agree with Alex's proposal (which more-or-less brings us back to the first commit I made to this PR.) I've updated it again and brought it back to the two-space model for input, which feels better and better the more I look at it.

The one thing from Alex's strawman that isn't present here is the `XRPose.ray` attribute, because we had previously defined that `XRRay` has a constructor which takes in a transform, and that seemed sufficient for now. The nice part is that if we identify a concrete need for the `.ray` attribute later it's a very easy, additive bit of API surface to put in.

input-explainer.md
```diff
@@ -92,10 +94,10 @@ function drawHighlightFrom(hoveredObject, inputSource) {
 function drawScene() {
   // Display only a single cursor or ray, on the most recently used input source.
   if (lastInputSource) {
-    let inputPose = xrFrame.getInputPose(lastInputSource, xrReferenceSpace);
-    if (inputPose) {
+    let targetRayPose = xrFrame.getPose(lastInputSource.targetRaySpace, xrReferenceSpace);
```
NellWaliczek: Unrelated to this PR...
`getInputSources()` should not include an entry for touch-screen input, yes?  Instead it creates a temporary `XRInputSource` so that it can fire the select event and pass it as a parameter.  What is the expected behavior if someone saves that `XRInputSource` and tries to use it with an XRFrame that's unsupported? I assume the `XRPose` would be null?
In either case, we probably should update the sample where lastInputSource is set, to encourage the right behavior.

input-explainer.md
```diff
@@ -173,37 +174,35 @@ function onSelect(event) {
...
+  * `'tracked-pointer'`: If the `gripSpace` is not `null` an application-appropriate controller model should be drawn using that transform. If appropriate for the experience, the a visualization of the target ray and a cursor as described in the `'gaze'` should also be drawn.
   * `'screen'`: In all cases the point of origin of the target ray is obvious and no visualization is needed.
 
 ```js
 // These methods presumes the use of a fictionalized rendering library.
 
 // Render a visualization of the input source - eg. a controller mesh.
-function renderInputSource(session, inputSource, inputPose) {
+function renderInputSource(session, inputSource, gripPose) {
   // FIXME: Using a fictional isDisplayOpaque() method to state that controller meshes should not be rendered
```
NellWaliczek: While you're in here, can we fix this to use the XREnvironmentBlendMode?

input-explainer.md
```diff
@@ -173,37 +174,35 @@ function onSelect(event) {
...
   // on transparent displays (AR).
-  if (isDisplayOpaque(session) && inputPose.gripTransform) {
-    // Render a controller mesh the using the gripTransform.
+  if (isDisplayOpaque(session) && gripPose) {
+    // Render a controller mesh if the using the gripPose as a transform.
```
NellWaliczek: I don't understand this sentence....?

input-explainer.md
```diff
@@ -231,21 +230,22 @@ function onSelectStart(event) {
   if (activeDragInteraction)
     return;
   
-  let inputPose = event.frame.getInputPose(event.inputSource, xrReferenceSpace);
   // Ignore the event if this input source is not capable of tracking.
-  if (!inputPose || !inputPose.gripTransform)
-    return;
+  if (!event.inputSource.gripSpace)
```
NellWaliczek: Wait.. why?

input-explainer.md
```diff
@@ -231,21 +230,22 @@ function onSelectStart(event) {
...
+  let targetRayPose = event.frame.getPose(event.inputSource.targetRaySpace, xrReferenceSpace);
+  let gripPose = event.frame.getPose(event.inputSource.gripSpace, xrReferenceSpace);
 
   // Use the input source target ray to find a draggable object in the scene
-  let ray = inputPose.targetRay;
-  let hitResult = scene.hitTest(ray.origin, ray.direction);
+  let hitResult = scene.hitTest(new XRRay(targetRayPose.transform));
   if (hitResult && hitResult.draggable) {
-    // Use the grip position to drag the intersected object, rather than the target ray.
+    // Use the gripPose position to drag the intersected object, rather than the target ray.
```
NellWaliczek: Yeah, I'm still not sure why this sample limits grabbing objects to input sources that have a grip pose...

input-explainer.md
```diff
@@ -259,19 +259,19 @@ function onSelectEnd(event) {
...
+      let deltaPosition = Vector3.subtract(gripPose.transform.position, activeDragInteraction.inputSourceStartPosition);
       let newPosition = Vector3.add(activeDragInteraction.targetStartPosition, deltaPosition);
       activeDragInteraction.target.setPosition(newPosition);
     }
   }
 }
 ```
 
-The above sample is optimized for dragging items in the scene around using input sources that have a `gripTransform`. It would also be possible to add further script logic to use the target ray properties to position items in the world - this is left as an exercise for the reader.
+The above sample is optimized for dragging items in the scene around using input sources that have a `gripSpace`. It would also be possible to add further script logic to use the target ray properties to position items in the world - this is left as an exercise for the reader.
```
NellWaliczek: Am I just being dense? Is this somehow harder?

input-explainer.md
```diff
@@ -25,33 +25,35 @@ The properties of an XRInputSource object are immutable. If a device can be mani
...
 ```js
 // Loop over every input source and get their pose for the current frame.
 for (let inputSource of xrInputSources) {
-  let inputPose = xrFrame.getInputPose(inputSource, xrReferenceSpace);
+  let targetRayPose = xrFrame.getPose(inputSource.targetRaySpace, xrReferenceSpace);
 
   // Check to see if the pose is valid
-  if (inputPose) {
+  if (targetRayPose) {
+    let gripPose = xrFrame.getPose(inputSource.gripSpace, xrFrameOfRef);
```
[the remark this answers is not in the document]
toji: Good catch, though I wonder if that's a bit of developer ergonomics that we should add? If we allow the first `XRSpace` argument of `getPose()` to be nullable if could potentially cut down on some redundant checks? Specifically, since `getPose()` can return `null`, we could simply have it do the error check for you and return `null` immediately if null is passed to allow a level of if statements to be stripped away. You'd see some code that could go from:

```js
if (inputSource.gripSpace) {
  let gripPose = xrFrame.getPose(inputSource.gripSpace, xrFrameOfRef);
  if (gripPose) {
    /* Do a thing. */
  }
}
```
to simply
```js
let gripPose = xrFrame.getPose(inputSource.gripSpace, xrFrameOfRef);
if (gripPose) {
  /* Do a thing. */
}
```

There's gonna be *some* cost to making the native call, though, so the most performance sensitive apps would still want to do a separate check. Hm...

input-explainer.md
toji: Huh... I could have sworn I did. Thanks for calling that out.

input-explainer.md
toji: Herp derp. Fixed.

input-explainer.md
toji: Hm.... it really shouldn't, now that you mention it. I'll work on that.

input-explainer.md
toji: TBH I don't know at this point either. Maybe something to do with the fact that it was an `XRRay` previously instead of a full transform? (Though given the logic shown here that seems like it would have actually made things easier, so...) In any case, I've refactored this to use `targetRaySpace` all the time now.

toji:
Made some updates based on your feedback.

input-explainer.md
```diff
@@ -25,33 +25,38 @@ The properties of an XRInputSource object are immutable. If a device can be mani
...
 for (let inputSource of xrInputSources) {
-  let inputPose = xrFrame.getInputPose(inputSource, xrReferenceSpace);
+  let targetRayPose = xrFrame.getPose(inputSource.targetRaySpace, xrReferenceSpace);
 
   // Check to see if the pose is valid
-  if (inputPose) {
-    // Render a visualization of the input source (see next section).
-    renderInputSource(session, inputSource, inputPose);
+  if (targetRayPose) {
+    if (inputSource.gripSpace) {
```
NellWaliczek: Drawing the motion controller shouldn't have anything to do with the viability of a targetRayPose.  In fact, I'd move lines 51-54 into `renderInputSource()` and just always call line 55 regardless of line 50.

spatial-tracking-explainer.md
```diff
@@ -349,24 +349,24 @@ partial interface XRSession {
...
+  XRPose? getPose(XRSpace space, XRSpace relativeTo);
 };
 
 [SecureContext, Exposed=Window]
-interface XRInputPose {
+interface XRPose {
   readonly attribute boolean emulatedPosition;
-  readonly attribute XRRay targetRay;
-  readonly attribute XRRigidTransform? gripTransform;
+  // TODO: Need for additional flags here?
```
NellWaliczek: Let's remove these TODOs for now and just include a sentence or two about them in the PR description when we move this to the WebXR repo.

input-explainer.md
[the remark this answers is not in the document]
toji: Resolved on call: We won't accept `null` for now.

input-explainer.md
```diff
@@ -25,33 +25,35 @@ The properties of an XRInputSource object are immutable. If a device can be mani
...
+The `gripSpace` represents a space where if the user was holding a straight rod in their hand it would be aligned with the negative Z axis (forward) and the origin rests at their palm. This enables developers to properly render a virtual object held in the user's hand. For example, a sword would be positioned so that the blade points directly down the negative Z axis and the center of the handle is at the origin.
 
-If the input source has only 3DOF, the `gripTransform` may represent only a translation or rotation based on tracking capability. An example of this case is for physical hands on some AR devices which only have a tracked position. The `gripTransform` will be `null` if the input source isn't trackable. 
+If the input source has only 3DOF, the grip pose may represent only a translation or rotation based on tracking capability. An example of this case is for physical hands on some AR devices which only have a tracked position. The `gripSpace` will be `null` if the input source isn't trackable.
```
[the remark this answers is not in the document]
thetuvix: I'm doubt any web devs will benefit from explicit handling of the corner case of a hand that delivers non-null position and null orientation.  Exactly 0 apps that don't explicitly test on a HoloLens will handle that case correctly, except for apps built on middleware that found that weird wart once and just paved over it at the middleware layer... at which point we surfaced that nuance from WebXR for nothing.

I'd expect Edge on HoloLens to just invent some hand orientation here, such as an Iron Man ray facing "forward" in the direction your head is facing.  This weird nuance can then be hidden from app developers entirely.

GIven that, I'd just define 3DOF more explicitly in WebXR as "orientation without position".

input-explainer.md
toji: Good call on moving the lines. Done.

toji:
Okay, made the changes discussed with the exception of the second flag, given Alex's latest comment. I'll wait for your approval prior to moving over to the webxr repo.

input-explainer.md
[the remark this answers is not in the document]
NellWaliczek: Ok, then :p

toji:
Moved to WebXR PR 460

PR #11, opened 2018-11-09
Teleportation + Transformation decomposition

toji:
This is a WIP, but I wanted to sanity check the direction with you prior to pursuing further.

Pretty confident about the actual mechanics of the frame of reference transform, though I'd appreciate if you can think of any problematic scenarios for it. I do feel that it probably justifies something a little more rigorous than just a `Float32Array` matrix, though, and that's where this starts to get messier, because if we introduce a ridged transform type then we probably ought to be using it everywhere. 😅

(I'll also admit that it's tough to start poking your head down this road without wanting to dive into coordinate system refactors ala #9.)

A few points worth calling out at this early stage:

 - Using `DOMPoint` to represent a Quaternion feels 😝 but it's got the right attributes for it and I really don't want to try and invent more geometric primitives for the web.
 - I'm not thrilled with the implied overhead of the additional objects and function call to get a `Float32Array`, but I'm starting to suspect that if most API usage is done via engines/libraries that decompose the transforms anyway this will actually end up being a performance win.
 - `XRProjection` type isn't really necessary here, but it feels profoundly weird to break one matrix down into a new interface and leave the other one as a pile of floats.
 - That said I'm not pleased with the inclusion of the `depthNear/Far` on the projection and want to handle it differently BUT I think we may also need to take that range into account during compositing in some cases so I don't think we can just make it into args passed into the `projection.toFloat32Matrix()` function (which would be nice and get rid of those values on the session.)


darktears:
As I commented in issue [431](https://github.com/immersive-web/webxr/issues/431) most of the time I had to decompose was to support the teleportation. Digging out Three.JS source code I couldn't find relevant occurrences where the engine needs to decompose in the critical path. But then even if my demo would not need to decompose anymore, I'm thinking that one common use case as a developer would probably to get the position in space (vr tracking + offset applied) so I'm thinking that maybe having the individual values would help.

Another thing to consider about decomposition; here is a typical usage in a Three.js application :
```javascript
    this._camera.projectionMatrix.fromArray(projectionMatrix);
    this._camera.matrixWorldInverse.copy(viewMatrix);
    this._scene.matrix.copy(viewMatrix);
```
It would seem that the float array is useful, however for the scene in Three.JS one has to do : 
```javascript
this._scene.matrixAutoUpdate = false;
//set the new matrix array as above.
this._scene.updateMatrixWorld(true);
```
when working directly with the matrix attribute. If the developer had the individual values, I'm almost sure one could operate directly on the [rotation](https://threejs.org/docs/#api/en/core/Object3D.rotation) and [position](https://threejs.org/docs/#api/en/core/Object3D.position) of the scene without messing the matrix auto update stuff (which you have to turn on again when exiting immersive mode to inline mode). 

The camera in Three.JS expects to work with matrices so the individual values are less useful in developer's code (but I haven't checked other engines like Babylon.JS for example). In Three.JS it would potentially be useful to simplify [this](https://github.com/mrdoob/three.js/blob/12450bedbef53649992cf55f10d7c1b9a253570f/src/renderers/webvr/WebVRUtils.js#L17).

toji:
@NellWaliczek: Updated this PR based on our conversation today, included everything but a Ray/Transform merger because I'm still not sure how I feel about that. PTAL when you can!

darktears:
@toji I like the latest changes, the removal of the toMatrix() looks cleaner.

toji:
Thanks Alexis! Credit to Nell for pushing for the attribute version.

darktears:
Thanks Nell :)

spatial-tracking-explainer.md
```diff
@@ -177,6 +177,35 @@ function onSessionStarted(session) {
 }
 ```
 
+### Application-supplied transforms
+Frequently developers will want to provide an additional, artificial transform on top of the user's tracked motion to allow the user to navigate larger scenes than their tracking systems or physical space allows. An examples of this would be the common "teleportation" mechanic, where the user selects a point in the virtual scene to "jump" to, after which the selected point is treated as the new origin which all tracked motion is relative to.
+
+While this effect can be accomplished by the application by mathematically combining the API-provided transform with the desired additional transforms, the same effect can also be achieved by setting the `baseTransform` attribute of any `XRReferenceSpace`. When the `transform` is not null, any values queried using the `XRReferenceSpace` will be offset by the `translation` and `orientation` the `baseTransform` describes. This is useful because it ensures that all tracked values are transformed consistently, allowing the user's viewpoint and inputs to stay in sync.
```
NellWaliczek: nit: pull the "why" to be the second sentence :)

spatial-tracking-explainer.md
```diff
@@ -177,6 +177,35 @@ function onSessionStarted(session) {
 }
 ```
 
+### Application-supplied transforms
```
NellWaliczek: Perhaps it's worthwhile to do this part as a separate PR when you move it to the main repo?

spatial-tracking-explainer.md
```diff
@@ -177,6 +177,35 @@ function onSessionStarted(session) {
...
+
+While this effect can be accomplished by the application by mathematically combining the API-provided transform with the desired additional transforms, the same effect can also be achieved by setting the `baseTransform` attribute of any `XRReferenceSpace`. When the `transform` is not null, any values queried using the `XRReferenceSpace` will be offset by the `translation` and `orientation` the `baseTransform` describes. This is useful because it ensures that all tracked values are transformed consistently, allowing the user's viewpoint and inputs to stay in sync.
+
+When interpreting an `XRRigidTransform` the `orientation` is applied prior to the `translation`. This means that, for example, a transform that indicates a quarter rotation to the right and a 1 meter translation along -Z would place a transformed object at `[0, 0, -1]` facing to the right.
+
+The following example demonstrates how to create a `floor-level` reference space to account for a scene where the virtual floor is modeled at 2 meters above the origin.
```
NellWaliczek: This example is a bit confusing to me... Can we do one that moves in X/Z instead?

spatial-tracking-explainer.md
```diff
@@ -177,6 +177,35 @@ function onSessionStarted(session) {
...
+    xrReferenceSpace.baseTransform = new XRRigidTransform({ y: 2 });
+  })
+  .then(setupWebGLLayer)
+  .then(() => {
+    xrSession.requestAnimationFrame(onDrawFrame);
+  });
+}
+```
+
+The `XRReferenceSpace`'s `baseTransform` can be updated at any time and will immediately take effect, meaning that any new values that are queried with the `XRReferenceSpace` will take into account the new `baseTransform`. Previously queried values will not be altered. Changing the `baseTransform` between pose queries in a single frame is not advised, since it will cause inconsistencies in the tracking data and rendered output.
```
NellWaliczek: Any concerns about external libraries making a mess of things by us not delaying the offset to the next frame boundary?

spatial-tracking-explainer.md
```diff
@@ -365,6 +394,8 @@ dictionary XRReferenceSpaceOptions {
 };
 
 [SecureContext, Exposed=Window] interface XRReferenceSpace : XRSpace {  
+  attribute XRRigidTransform? baseTransform;
```
NellWaliczek: nit: I'm not super thrilled with this name...it didn't feel very meaningful to me when I first read it... happy to be overruled, but maybe something like `offset` or `virtualOffset` would be better?  Dunno, not sure those are great either.

spatial-tracking-explainer.md
```diff
@@ -412,11 +443,11 @@ interface XRUnboundedReferenceSpace : XRReferenceSpace {
  Constructor(DOMString type, XRReferenceSpaceEventInit eventInitDict)]
 interface XRReferenceSpaceEvent : Event {
   readonly attribute XRReferenceSpace referenceSpace;
-  readonly attribute Float32Array? transformMatrix;
+  readonly attribute XRRigidTransform? transform;
```
NellWaliczek: nit: again.... this name isn't super meaningful (and yes, I know I named the original property ^_^)... i don't know what would be better though...

spatial-tracking-explainer.md
```diff
@@ -233,7 +262,7 @@ Additionally, XR hardware with orientation-only tracking may also provide an emu
...
 
 #### Inline to Immersive
-It is expected that developers will often choose to preview `immersive` experiences with a similar experience `inline`.  In this situation, users often expect to see the scene from the same perspective when they make the transition from `inline` to `immersive`. To accomplish this, developers should grab the `poseMatrix` of the last `XRViewerPose` retrieved using the `inline` session's `XRReferenceSpace` and apply this same transform content to be displayed in the `immersive` session's `XRReferenceSpace`.  The same logic applies in the reverse when exiting `immersive`.
+It is expected that developers will often choose to preview `immersive` experiences with a similar experience `inline`.  In this situation, users often expect to see the scene from the same perspective when they make the transition from `inline` to `immersive`. To accomplish this, developers should grab the `transform` of the last `XRViewerPose` retrieved using the `inline` session's `XRReferenceSpace` and set it as the `transform` of the `immersive` session's `XRReferenceSpace`.  The same logic applies in the reverse when exiting `immersive`.
```
NellWaliczek: Assuming the name `baseTransform` stays (see my comment below), shouldn't this be...

>set it as the `baseTransform` of the `immersive` session's `XRReferenceSpace`.

input-explainer.md
```diff
@@ -161,8 +161,8 @@ function onSelect(event) {
 
   // Ray cast into scene with the viewer pose to determine if anything was hit.
   // Assumes the use of a fictionalized math and scene library.
-  let rayOrigin = getTranslation(viewerPose.poseMatrix);
-  let rayDirection = applyRotation(scene.forwardVector, viewerPose.poseMatrix);
+  let rayOrigin = viewerPose.transform.translation;
+  let rayDirection = applyRotation(scene.forwardVector, viewerPose.transform.orientation);
```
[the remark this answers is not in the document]
toji: Completely agree. 👍

spatial-tracking-explainer.md
toji: I'm not necessarily a bit fan of this name either, but I shied away from `offset` or any derivatives because that makes it sounds like a translation-only transform, which it's not. I could be convinced to go with it, though, if everyone else feels like it doesn't carry that connotation.

In general, I'm completely open to a better name.

darktears: IMHO `transform` is the right term because it carries the idea of translation + rotation and it's a well established term for 3D developers. Maybe transformOrigin since it's going to be the origin in which the XR system transforms are going to be applied?

spatial-tracking-explainer.md
toji: We could, but most of the time my concern about delays like that are simply ensuring that the state syncs appropriately with the backing native API. In this case we know we can handle the attribute purely via mathematics, so I didn't see the need for a delay.

explainer.md
```diff
@@ -607,14 +607,23 @@ enum XREye {
   "right"
 };
 
+[SecureContext, Exposed=Window,
+ Constructor(optional DOMPointInit translation, optional DOMPointInit orientation)]
```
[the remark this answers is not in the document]
toji: Not sure how I'm feeling about the matrix constructor. Fortunately that's really easy to add later as an additive improvement if it's needed!

spatial-tracking-explainer.md
toji: That's doable. (The addition of XRRigidTransform feels like it lacks justification on it's own, but if we know this is in the pipeline it's no big deal.)

spatial-tracking-explainer.md
toji: 🤷‍♀️

spatial-tracking-explainer.md
toji: `transformOrigin` or some variant thereof doesn't feel bad to me.

toji:
Updated the PR to address the feedback so far.

spatial-tracking-explainer.md
toji: Jumped through some rebase hoops and worked this PR down to two commits, one that just introduces `XRRigidTransform` and one that adds the teleport logic. When we push this to the main repo I'll push each commit as a separate PR.

toji:
Moved to WebXR repo PRs 440 and 441 to facilitate wider WG discussion. (Didn't make a name change to the `XRReferenceSpace` transform yet, still open to `transformOrigin`).

explainer.md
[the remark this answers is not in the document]
NellWaliczek: Yeah, I have a suspicion it would be helpful for at least Sumerian, but we can wait until i have that suspicion confirmed :)

spatial-tracking-explainer.md
NellWaliczek: Ok, I buy that.

PR #13, opened 2018-11-19
More explicitly describe frame lifetime as related to XRSession state.

toji:
This isn't something there's an issue open on directly at the moment, but it's a topic that's come up a lot in various conversations about how to process values that may have an effect on frame processing. (Rafael has been persistent about bringing this problem up).

The basic problem is this: If I call `requestAnimationFrame()`, then wait a while before changing, say, the `baseLayer` prior to the frame callback firing, what happens? For the most part we can imagine that doing so should be fine, and that we could process the layer contents without issue even with the last second swap, but what if the layer had information on it that needed to be fed into the frame request itself? This gets more problematic as the session state gets more complicated, which is certain to happen. (See issues like "Default FoV for magic window canvases" and the desire to reduce complexity of the `outputContext`, which will likely make it mutable, not to mention allowing multiple layers.)

So this just adds a paragraph to the explainer that sets out very precisely when changes to `XRSession` state are picked up for a frame to avoid having this conversation over and over again for every new property. (Proposal TL;DR: state is snapshotted at the first rAF call for every new frame cycle.)

I'm also toying with an idea for a follow-up to this that would bundle all presentation state into a single object on the session to make it clearer what is picked up on the frame boundaries (and maybe make multi-layer support cleaner). Something like:

```js
xrSession.setPresentationState({
  depthNear: 1,
  depthFar: 1000,
  outputContext: someCanvas,
  layers: [ webglLayer ]
});

onFrame(timestamp, frame) {
  let baseLayer = frame.presentationState.layers[0];
  /* ... */
}
```

¯\_(ツ)_/¯


NellWaliczek:
Sooooo way back when I did the initial proposals for what became the XRFrame and the UA-provided buffers I had the same sort of concern... ;) I'm on board with the premise of this change.  Assuming this is the approach we take I have two bikeshedding-style concerns.
1. For now I'd say the `layers` attribute should just be `baseLayer` for consistency with the current design (though yes, I'm aware my opening the XRQuadLayer can of worms will probably change that, but one thing at a time!~)
2. Not sure how I feel about the term `presentation` ... I wouldn't want to give the impression that depth near/far and baseLayer were only for the associated `XRPresentationContext`.

Which actually brings me to my other question... which is... what about the `requestViewportScaling()` call?  That's on the `XRWebGLLayer` directly but is another one that shouldn't apply until the next frame, yeah?  And maybe more importantly, is that the only functionality on the `XRWebGLLayer` that will ever behave that way?

toji:
The viewport question is a good one, and one that I was thinking about too, though I didn't want to over-complicate things by diving into it. I definitely don't think it's practical to move the viewport scaling into session state, since it may not apply to all layers and you may want different scaling for different layers even if it does.

The most reasonable logic I've come to is to simply say that the frame state includes the state of all bound layers at the time the frame was requested. I'm not sure if viewport scale is the only property that will need that treatment or not, but either way that framing would lead to consistent behavior.

(We could reinforce it a little by having a similar state-object-set-as-a-dictionary on the layers too so that there's at least an ergonomic callback between the two, but that seems like significant overkill for only one or two properties.)

toji:
Updated the explainer text to hopefully resolve your feedback and moved the PR to the webxr repo (PR 439). Broke the conversation about encapsulating the state in question in a separate interface out into a different PR, #17.

PR #17, opened 2018-12-07
Encapsulate session/frame render state into it's own interface

toji:
WIP, not ready for migration to webxr repo yet.

Questions I have so far:

- "XRRenderState"? 👍 or 👎 
- Should updating the state be required to be a full update each call? So if I want to update the depth range do I also have to pass in the layers again, or should it only update the dictionary entries that are specified? Full update is cleaner from a code/spec perspective but could be a pain for developers.
- Expose active `renderState` on the session, frame, or both?


explainer.md
```diff
@@ -134,7 +134,7 @@ Only one immersive session per XR hardware device is allowed at a time across th
 
 Once the session has started, some setup must be done to prepare for rendering.
 - An `XRReferenceSpace` should be created to establish a space in which `XRViewerPose` data will be defined. See the [Spatial Tracking Explainer](spatial-tracking-explainer.md) for more information.
-- An `XRLayer` must be created and assigned to the `XRSession`'s `baseLayer` attribute. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
+- An `XRLayer` must be created and assigned to the `XRSession`'s `renderState.baseLayer` attribute. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
```
NellWaliczek: Isn't this a promise now?

explainer.md
```diff
@@ -251,6 +256,10 @@ function drawScene(view) {
 }
 ```
 
+### Frame lifetime
```
NellWaliczek: ?

explainer.md
```diff
@@ -564,16 +578,14 @@ dictionary XRSessionCreationOptions {
   readonly attribute XRSessionMode mode;
   readonly attribute XRPresentationContext outputContext;
```
NellWaliczek: Is there a reason you decided to leave this off for now?

explainer.md
```diff
@@ -564,16 +578,14 @@ dictionary XRSessionCreationOptions {
...
-  attribute double depthFar;
-
-  attribute XRLayer baseLayer;
+  readonly attribute XRRenderState renderState;
 
   attribute EventHandler onblur;
   attribute EventHandler onfocus;
   attribute EventHandler onend;
 
+  Promise<void> setRenderState(XRRenderStateInit state);
```
NellWaliczek: nit: What about `setRenderOptions`?  The word "state" feels more to me like something that needs to be done every frame.

explainer.md
```diff
@@ -590,6 +602,18 @@ enum XREnvironmentBlendMode {
   "alpha-blend",
 };
 
+dictionary XRRenderStateInit {
+  double depthNear = 0.1;
```
NellWaliczek: I'd prefer if developers could choose to leave off values they didn't want to set without overwriting changes they may have made previously.

NellWaliczek:
I think I addressed everything but
>Expose active renderState on the session, frame, or both?

And... ummmm I dunno?

explainer.md
toji: As we covered on the most recent editors sync call, I definitely feel like this SHOULD be part of the render state, but I'd like to handle it as a separate PR because making it mutable requires a bit more logic that I'd prefer to not clutter this PR with.

explainer.md
toji: I'm not sure I feel the same way. I guess I think about it in the context of OpenGL/WebGL context state, where it's whatever you last set it as regardless of how long ago you did it.

toji:
Okay, updated this PR to address feedback. Also tweaked a couple of things, such as renaming the function to `updateRenderState` to account for the fact that it's no longer doing a full state replacement.

>> Expose active renderState on the session, frame, or both?

> And... ummmm I dunno?

Leaving it on `XRSession` for now because that's always available from the `XRFrame` and there's only going to be one active `XRFrame` at a time, as we discussed on the call.

explainer.md
toji: Made this change, and also reached out internally to see if that's a weird thing from an ergo perspective.

explainer.md
toji: Rebase issue. Removed.

explainer.md
toji: Yup, reworked the text.

explainer.md
```diff
@@ -134,7 +134,7 @@ Only one immersive session per XR hardware device is allowed at a time across th
 
 Once the session has started, some setup must be done to prepare for rendering.
 - An `XRReferenceSpace` should be created to establish a space in which `XRViewerPose` data will be defined. See the [Spatial Tracking Explainer](spatial-tracking-explainer.md) for more information.
-- An `XRLayer` must be created and assigned to the `XRSession`'s `baseLayer` attribute. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
+- An `XRLayer` must be created and set as the `XRSession`'s `renderState` `baseLayer`. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
```
NellWaliczek: `renderState.baseLayer`?

explainer.md
```diff
@@ -159,7 +159,9 @@ function onSessionStarted(session) {
 
 ### Setting up an XRLayer
 
-The content to present to the device is defined by an `XRLayer`. In the initial version of the spec only one layer type, `XRWebGLLayer`, is defined and only one layer can be used at a time. This is set via the `XRSession`'s `baseLayer` attribute. Future iterations of the spec will define new types of `XRLayer`s. For example: a new layer type would be added to enable use with any new graphics APIs that get added to the browser. The ability to use multiple layers at once and have them composited by the UA will likely also be added in a future API revision.
+The content to present to the device is defined by an `XRLayer`. In the initial version of the spec only one layer type, `XRWebGLLayer`, is defined and only one layer can be used at a time. This is set via the `XRSession`'s `updateRenderState()` function. `updateRenderState()` takes a dictionary containing new values for a variety of options affecting the session's rendering, including `baseLayer`. Only the options specified in the dictionary are updated, and a promise is returned that resolves when the new values take effect.
```
NellWaliczek: >variety of options affecting the session's rendering

See, this is why I want to call it `updateRenderOptions()` ;)

explainer.md
```diff
@@ -178,7 +180,7 @@ function setupWebGLLayer() {
   return gl.makeXRCompatible().then(() => {
     // The content that will be shown on the device is defined by the session's
     // baseLayer.
-    xrSession.baseLayer = new XRWebGLLayer(xrSession, gl);
+    xrSession.updateRenderState({ baseLayer: new XRWebGLLayer(xrSession, gl) });
```
NellWaliczek: Should the sample code register for the animation frame callback in a .then?

toji: Not in this case because the `setupWebGLLayer()` is shown being called in a previous code sample where `session.rAF` is called in a `.then()` on the return value of `setupWebGLLayer()`. That ordering is weird when viewed in isolation, but helps support the explainers narrative flow.

NellWaliczek: I think there may be too many promises going on here... Because this isn't waiting until updateRenderState is finished, it's just waiting until makeXRCompatible is finished... right?

toji: Ah, I see what you meant now. Yeah, I missed that. I've updated it to return the promise from `updateRenderState()`.

toji:
Okay, took the latest feedback into account and posted to the WebXR repo as PR 451.

Issue #15, opened 2018-11-30
Discussion: Enabling motion tracking for inline sessions.


toji:
We need to have a discussion about exactly how we want to have developers indicate the need for motion tracking on an inline session. I think the assumption has been that the creation of a stationary reference space would be the signal that the developer wants to engage in any available tracking, but I'm concerned about how that interacts with the rest of our user consent flow. Specifically, if enabling tracking requires any sort of user consent and the creation of the reference space is what triggers it you may inadvertently end up in a situation where you're presenting multiple dialogs to the user (some during session creation and some during reference space creation). It may also imply that the space creation be gated on user activation so that we're not showing users consent dialogs at random. (That could be worked around by having the consent request be shown inline on the canvas or something, but that's a UA descision.) Additionally, it would be a break in behavior between immersive sessions, which assume that you can create reference spaces without user input of any sort, and requires a divergent code path for inline which we generally want to reduce.

It seems to me like a reasonable alternative would be to say that inline sessions get the same guarantees about stationary reference space availability that other session types get, but those reference spaces are inert unless you've requested an "inline-tracking" feature (or similar) at session creation time. That would appear to mesh with the immersive use cases better and you get an added benefit of not needing to do additional transform math on the viewer poses just to get them to line up with a floor-level scene just because you didn't actually need tracking.

I'm sure I'm overlooking something in my overly-simplistic assessment here, which is why I'd like to discuss it prior to making any actual proposals.

NellWaliczek:
Yeah, I very strongly agree that the creation of an `XRReferenceSpace` should not ever pop UI.  

And I'd also been assuming that we'd address this issue via a feature request, and is a big part of why I'm not super comfortable with treating `XRReferenceSpace`s as different in the desired/required pattern.  For example, I'd like to see developers decide for themselves if motion tracking is a desired features (they'll support mouse/keyboard/touch method for changing the viewer's location) or required.

In other words, I don't think the feature that should be requested is `inline-tracking` but instead it should be `stationary-reference-space`.

cwilso:
Hey, shouldn't this issue be over in the core spec issues?

toji:
I also felt there was a decent chance that I had just misunderstood an element of the spatial tracking work that had already been written, so I wanted to sanity check with Nell prior to making noise on the core issue list. I'm perfectly happy to create issues on the webxr repo once I've established what's actually missing. :D

toji:
[...]

PR #18, opened 2018-12-27
Describe the identity reference space

toji:
Shouldn't be anything too surprising in here, though there are a couple of things I wanted to call out to get your opinion on. (I'll put them in inline comments)


explainer.md
```diff
@@ -366,7 +366,7 @@ In order to mirror WebXR content to the page, developers provide an `XRPresentat
 
 When mirroring only one eye's content will be shown, and it should be shown without any distortion to correct for headset optics. The UA may choose to crop the image shown, display it at a lower resolution than originally rendered, and the mirror may be multiple frames behind the image shown in the headset. The mirror may include or exclude elements added by the underlying XR system (such as visualizations of room boundaries) at the UA's discretion. Pages should not rely on a particular timing or presentation of mirrored content, it's really just for the benefit of bystanders or demo operators.
 
-The UA may also choose to ignore the `outputCanvas` on systems where mirroring is inappropriate, such as devices without an external display to mirror to like mobile or all-in-one systems.
+The UA may also choose to ignore the `outputContext` on systems where mirroring is inappropriate, such as devices without an external display to mirror to like mobile or all-in-one systems.
```
toji: Just a typo fix. :)

spatial-tracking-explainer.md
```diff
@@ -196,12 +196,63 @@ function teleport(deltaX, deltaY, deltaZ) {
 }
 ```
 
+### Identity reference spaces
```
toji: What do you think about the name `identity`? Feels appropriate, but I could see going with a couple of other options as well.

spatial-tracking-explainer.md
```diff
@@ -196,12 +196,63 @@ function teleport(deltaX, deltaY, deltaZ) {
...
+An _identity_ reference space is one which provides no tracking. This type of reference space is used for creating inline experiences with tracing information explicitly disabled, such as a furniture viewer that will use pointer data to rotate the furniture. It also supports cases where the developer does not wish to request consent to access tracking data from the user prior to entering an immersive mode.
+
+This type of reference space is requested with a type of `identity` and returns a basic `XRReferenceSpace`. `XRViewerPose` objects retrieved with this reference space will have a transform that is equal to the reference space's `originOffset`. The pose will only contain a single `XRView`, which must also have a transform equal to the `originOffset`. `identity` references spaces cannot be created by immersive sessions, since a lack of tracking in an immersive experience could cause severe discomfort for the user.
```
toji: > requested with a type of `identity` and returns a basic `XRReferenceSpace`

Does this sit well with you? Given that this type doesn't need any additional members, it felt kinda silly to create a whole new (empty) interface for it. Doesn't seem like there's any harm in just returning a base `XRReferenceSpace` object for this one.

spatial-tracking-explainer.md
```diff
@@ -196,12 +196,63 @@ function teleport(deltaX, deltaY, deltaZ) {
...
+Inline sessions, by definition, do not require a user gesture or user permission to create, and as a result there must be strong limitations on the pose data that can be reported for privacy and security reasons.  Requests for an `XRBoundedReferenceSpace` or an `XRUnboundedReferenceSpace` will always be rejected on inline sessions, while requests for `identity` reference spaces will always succeed. Requests for an `XRStationaryReferenceSpace` may succeed, but may also be rejected if the UA is unable provide any tracking information such as for an inline session on a desktop PC or a 2D browser window in a headset. The UA is also allowed to request the users consent prior to returning an `XRStationaryReferenceSpace`.
+
+### Click-and-drag view controls
```
toji: This example seemed appropriate, as it illustrates a reasonable way to make use of the identity reference space, but it's also kinda disconnected from it here because it could also feasibly be used to good effect on a stationary reference space, which makes it feel a bit out-of-place in this PR. 🤷‍♀️

spatial-tracking-explainer.md
```diff
@@ -302,13 +353,14 @@ How to pick a reference space:
...
-| Type                           | Subtype             | Examples                                      |
-| -------------------------------| ------------------- | --------------------------------------------- |
-| `XRStationaryReferenceSpace` | `position-disabled` | - 360 photo/video viewer |
-| `XRStationaryReferenceSpace` | `eye-level`         | - Immersive 2D video viewer<br>- Racing simulator<br>- Solar system explorer |
-| `XRStationaryReferenceSpace` | `floor-level`       | - VR chat "room"<br>- Action game where you duck and dodge in place<br>- Fallback for Bounded experience that relies on teleportation instead |
-| `XRBoundedReferenceSpace`    |                     | - VR painting/sculpting tool<br>- Training simulators<br>- Dance games<br>- Previewing of 3D objects in the real world |
-| `XRUnboundedReferenceSpace`  |                     | - Campus tour<br>- Renovation preview |
+| Type         | Subtype             | Examples                                      |
+| -------------| ------------------- | --------------------------------------------- |
+| `identity`   |                     | - In-page content preview<br>- Click/Drag viewing |
```
toji: Slightly awkward side-effect of using the `XRReferenceSpace` as the return type for `identity` directly is that it's no longer ideal to refer to the types by their interface name here. IMO it kinda reads better this way, though, so I don't see that as a big loss.

toji:
Ooh, and I forgot to note that this addresses #15, which I'm going to close now since we have a PR tracking it.

spatial-tracking-explainer.md
NellWaliczek: I like it!

spatial-tracking-explainer.md
```diff
@@ -196,12 +196,63 @@ function teleport(deltaX, deltaY, deltaZ) {
 }
 ```
 
+### Identity reference spaces
+An _identity_ reference space is one which provides no tracking. This type of reference space is used for creating inline experiences with tracing information explicitly disabled, such as a furniture viewer that will use pointer data to rotate the furniture. It also supports cases where the developer does not wish to request consent to access tracking data from the user prior to entering an immersive mode.
```
NellWaliczek: nit: tracing = tracking

NellWaliczek: > It also supports cases where the developer does not wish to request consent to access tracking data from the user prior to entering an immersive mode.

I'm not really sure what you mean by this...

spatial-tracking-explainer.md
NellWaliczek: Yeah, I think that's probably fine.  The only concerns would be:
* If we want to add attributes to the base that are meaningless for the identity type, but that'd be a problem anyway.
* If we want to add attributes to the identity type, but I think adding a new subtype later shouldn't be a breaking change, yeah?

NellWaliczek: I'd suggest linking to the new "Click-and-drag view controls" section here because without that part, this new type is pretty meaningless ;)

spatial-tracking-explainer.md
```diff
@@ -196,12 +196,63 @@ function teleport(deltaX, deltaY, deltaZ) {
...
+Inline sessions, by definition, do not require a user gesture or user permission to create, and as a result there must be strong limitations on the pose data that can be reported for privacy and security reasons.  Requests for an `XRBoundedReferenceSpace` or an `XRUnboundedReferenceSpace` will always be rejected on inline sessions, while requests for `identity` reference spaces will always succeed. Requests for an `XRStationaryReferenceSpace` may succeed, but may also be rejected if the UA is unable provide any tracking information such as for an inline session on a desktop PC or a 2D browser window in a headset. The UA is also allowed to request the users consent prior to returning an `XRStationaryReferenceSpace`.
+
+### Click-and-drag view controls
+
+Frequently with inline sessions it's desirable to have the view rotate when the user interacts with the inline canvas. This is useful on devices without tracking capabilities to allow users to still view the full immersive scene, but can also be desirable on devices with tracking capabilities as a way to adjust the users view without requiring them to physically turn around.
```
NellWaliczek: >the full immersive scene

Isn't this just for inline?

NellWaliczek: Give an example of the type of device this would be helpful on (aka. a tablet or phone)

spatial-tracking-explainer.md
NellWaliczek: Gosh I'm torn by this.   I'd like all of our new prose/samples in the explainer to be almost copy&paste into the spec.  So, while I think having a sample like this is useful, it doesn't seem appropriate to the spec because it's really an developer choice.  Got any ideas on how we might illustrate the concept without actually putting this math in the spec?

As for whether it goes in this PR.. Eh, it's not a signature or functionality change, so I'm not super concerned about the PRs getting conflated.

spatial-tracking-explainer.md
NellWaliczek: Yeah, fine by me 👍

spatial-tracking-explainer.md
toji: Right. If we found that for some reason we wanted to add features only to the identity type (I can't imagine what) we would add a new interface, `XRIdentityReferenceSpace`, that extends the `XRReferenceSpace` and thus would be functionally identical except for the new stuff. If the developer had been checking the type previously using `instanceof` (like they should) then it will continue to correctly identify the new type as an instance of an `XRReferenceSpace`, so no breakage there. You might get screwed up in some very rare cases where the developer is doing something silly like

```js
if (Object.getPrototypeOf(identityRefSpace) == XRReferenceSpace.prototype) { /* ... * }
```

But, umm.... I kinda feel like you deserve to get broken if your code is depending on that? (And, again, I don't anticipate it actually happening.)

spatial-tracking-explainer.md
toji: Yes, it is. This is just a mental namespace conflict. Thanks for catching!

spatial-tracking-explainer.md
toji: Yeah, that's awkward. Sorry. Adjusted the wording on this to make it clearer.

toji:
Thanks for the comments! Addressed your feedback, please take another look.

spatial-tracking-explainer.md
NellWaliczek: Maybe a question to raise when we post the final PR to the main repo?

toji:
SGTM! Thanks! I'll work on moving it over, though the fact that I originally did it as a commit on top of #9 will make that job ever so slightly harder. #DidItToMyself

NellWaliczek:
I think #9 is really close if you want to wait a day or two...

toji:
Moved to WebXR PR 459

PR #19, opened 2019-01-04
Anchors and hit testing

NellWaliczek:
Really early thinking about how to bring in hit-testing and anchors on top of the pose refactor we just completed.  As I mentioned earlier, I really think the whole input explainer needs to be cleaned up because concepts are scattered around the document.
Also, because I based this branch on the PR in the webxr repo, when you review the changes, make sure you're only looking at the commits I added on top.

### DIFF
--- a/input-explainer.md
+++ b/input-explainer.md
@@ -295,10 +295,6 @@ function onSessionStarted(session) {
   // Rest of session start up logic
 }
 
-
-  session.addEventListener("inputsourceschange", ev => {
-  });
-
 let preferredHitTestSource = null;
 function onInputSourcesChange(event) {
   let oldPreferredInputSource = preferredInputSource;
--- a/input-explainer.md
+++ b/input-explainer.md
@@ -395,25 +395,49 @@ interface XRRay {
   readonly attribute Float32Array matrix;
 };
 
+[SecureContext, Exposed=Window]
+interface XRInputSource {
+  readonly attribute XRSpace targetRaySpace;
+  readonly attribute XRHitTestSource hitTestSource;
+};
+
 enum XRHandedness {
   "",
   "left",
   "right"
 };
 
-enum XRTargetRayMode {
-  "gaze",
-  "tracked-pointer",
-  "screen"
+[SecureContext, Exposed=Window]
+interface XRTrackedInputSource : XRInputSource {
+  readonly attribute XRHandedness handedness;
+  readonly attribute XRSpace gripSpace;
+
+  // From the other PR
+  readonly attribute XRTrackedController? controller;
+  readonly attribute DOMString renderId;
 };
 
+// Per our conversation, I'm contemplating the approach that one of these should
+// ALWAYS be present with gamepad=null.  One and only one.  Regardless of whether 
+// or not there's other input sources.  It's a tad bit weird if there's a gamepad
+// plugged in, but I think it's less confusing to have two XRViewerInputSource
+// objects (one with gamepad and one without) to ensure that the oninputsourcechange 
+// event behaves as expected should the gamepad be unplugged.
+//
+// The bonus to this approach is that in the event handler for select, developers can
+// query the button on the gamepad that actually fired the event...
+//
+// Also your point about tracked controllers potentially having both hands and buttons
+// can also  be applied to the XRViewerInputSource if needed.
 [SecureContext, Exposed=Window]
-interface XRInputSource {
-  readonly attribute XRHandedness handedness;
-  readonly attribute XRTargetRayMode targetRayMode;
-  readonly attribute XRSpace targetRaySpace;
-  readonly attribute XRSpace? gripSpace;
-  readonly attribute XRHitTestSource hitTestSource;
+interface XRViewerInputSource : XRInputSource {
+  readonly attribute Gamepad? gamepad;
+};
+
+// One per touch-point
+[SecureContext, Exposed=Window]
+interface XRScreenInputSource : XRInputSource {
+  // Add screen coordinates?
 };
 
 //
--- a/input-explainer.md
+++ b/input-explainer.md
@@ -56,7 +56,7 @@ for (let inputSource of xrInputSources) {
     }
   }
 
-  // Render a visualization of the input source if appropriate.              
+  // Render a visualization of the input source if appropriate.
   // (See next section for details).
   renderInputSource(xrFrame, inputSource);
 }
@@ -283,6 +283,7 @@ For each of these events the `XRInputSource`'s target ray must be updated to ori
 ### Environmental hit-testing
 Descriptive text goes here
 
+Basic example of using input sources to get hit test results each frame
 ```js
 function onDrawFrame(timestamp, xrFrame) {
   // Frame logic
@@ -293,12 +294,10 @@ function onDrawFrame(timestamp, xrFrame) {
   // Other frame logic
 }
 ```
-```js
-let objectBounds = { frontLeft: new XRRigidTransform({x:-.5, y:0, z:.5),
-                     frontRight: new XRRigidTransform({x:.5, y:0, z:.5),
-                     backLeft: new XRRigidTransform({x:-.5, y:0, z:-.5),
-                     backRight: new XRRigidTransform({x:.5, y:0, z:-.5) };
 
+Sample of creating a secondary hit test source to find valid planes for dropping an object.
+This code is cribbed from a section above and then updated
+```js
 function onSelectStart(event) {
   // Ignore the event if we are already dragging
   if (activeDragInteraction)
@@ -315,36 +314,50 @@ function onSelectStart(event) {
       targetStartPosition: hitResult.position,
       inputSource: event.inputSource,
       inputSourceStartPosition: targetRayPose.transform.position;
-      hitTestBounds: {};
     };
-  }
 
-  foreach (key in objectBounds.keys) {
-    xrSession.requestHitTestSource(event.inputSource.targetRaySpace, objectBounds[key]).then((hitTestSource) => {
-      activeDragInteraction.hitTestBounds[key] = hitTestSource;
+    xrSession.requestHitTestSource({ space:event.inputSource.targetRaySpace, type:"horizontal-plane"}).then((hitTestSource) => {
+      activeDragInteraction.hitTestSource = hitTestSource;
     });
+  }
+}
+
+let anchors = [];
+
+// Only end the drag when the input source that started dragging releases the select action
+function onSelectEnd(event) {
+  if (activeDragInteraction && event.inputSource == activeDragInteraction.inputSource) {
+    let frame = event.frame;
+    let virtualHitResult = scene.hitTest(new XRRay(targetRayPose.transform));
+    if (activeDragInteraction.hitTestSource) {
+      let environmentHitTestResult = frame.getHitTestResults(activeDragInteraction.hitTestSource, xrReferenceSpace);
+    }
+
+    let targetPose = scene.getUpdatedDragTarget(virtualHitResult, environmentalHitResult);
+    if (targetpose != null && targetPose != virtualHitTestResult) {
+      frame.requestAnchor({space:xrReferenceSpace, offset:targetPose, type:activeDragInteraction.hitTestSource.type}).then((anchor) => {
+        anchors.add(anchor);
+      });
+    }
+
+    activeDragInteraction = null;
   }
 }
 
 function onUpdateScene() {
   if (activeDragInteraction) {
-    let targetPose = null;
     
-    let environmentHitTestResults = xrFrame.getHitTestResults(preferredInputSource.hitTestSource, xrReferenceSpace);
-    
-    let boundsHitTestResults = {};
-    foreach (key in activeDragInteraction.hitTestBounds) {
-      boundsHitTestResults[key] = xrFrame.getHitTestResults(activeDragInteraction.hitTestBounds, xrReferenceSpace);
+    let virtualHitResult = scene.hitTest(new XRRay(targetRayPose.transform));
+    if (activeDragInteraction.hitTestSource) {
+      let environmentHitTestResult = xrFrame.getHitTestResults(activeDragInteraction.hitTestSource, xrReferenceSpace);
     }
 
-    let hitResult = scene.hitTest(new XRRay(targetRayPose.transform));
-
-    // TODO some logic to make the determination that the spot is valid and set targetPose
+    let targetPose = scene.getUpdatedDragTarget(virtualHitResult, environmentalHitResult);
 
     if (targetPose) {
       // Determine the vector from the start of the drag to the input source's current position
       // and position the draggable object accordingly
-      let deltaPosition = Vector3.subtract(transform.position, activeDragInteraction.inputSourceStartPosition);
+      let deltaPosition = Vector3.subtract(targetPose.position, activeDragInteraction.inputSourceStartPosition);
       let newPosition = Vector3.add(activeDragInteraction.targetStartPosition, deltaPosition);
       activeDragInteraction.target.setPosition(newPosition); // TODO make this an XRRigidTransform function?
     }
@@ -360,7 +373,7 @@ This is a partial IDL and is considered additive to the core IDL found in the ma
 //
 
 partial interface XRSession {
-  Promise<XRHitTestSource> requestHitTestSource(XRSpace space, optional XRRay ray);
+  Promise<XRHitTestSource> requestHitTestSource(XRSpace space, optional XRRay ray, optional XRHitTestType hitTestType);
 
   FrozenArray<XRInputSource> getInputSources();
   
@@ -397,7 +410,12 @@ interface XRRay {
 
 [SecureContext, Exposed=Window]
 interface XRInputSource {
-  readonly attribute XRSpace targetRaySpace;
+  // TODO: Consider if we should we remove this in favor just using hitTestSource.InputSource?
+  readonly attribute XRSpace targetRaySpace; 
+
+  // hitTestSource.space = targetRaySpace
+  // hittestSource.type = ""
+  // hitTestSource.offsetRay = FORWARD RAY
   readonly attribute XRHitTestSource hitTestSource;
 };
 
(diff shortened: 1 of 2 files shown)


NellWaliczek:
i've got a handful of concerns about specifics, but i'm feeling good about the general approach
1. i'm thinking about the timing difference in how long it takes to register a hittest source and whether or not it can result in the preferredHitTestSource being preempted
2. i'm a bit concerned it isn;t clear that registering a hit test source means it will continue to update as the space moves.  the function is on session, so maybe that's ok
3. I'm wondering if this design has a clear way to solve the problem of "give me a hit test that will fit the size of this virtual object so I can put it there".  Perhaps we can add a parameter for that to the hit test call itself?  However, I don't see a good way find out how big a virtual thing will fit.  Maybe an extra value on the hittest result?
4. I go back and forth on whether an `XRHitTestSource` should be a space itself... thoughts?
5. Should XRAnchors exist or should it just be XRSpace for now?

NellWaliczek:
Another key question is: how much of a concern is it that screen-based input would have several frames of latency before the hit test would begin reporting (though it would not have subsequent latency)

toji:
FYI: Re-uploaded the pose-refactor branch and made it the base for this to aid readability.

input-explainer.md
```diff
@@ -280,6 +280,57 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+  // Rest of session start up logic
+}
+
+let preferredHitTestSource = null;
+function onInputSourcesChange(event) {
+  let oldPreferredInputSource = preferredInputSource;
+
+  // Choose an appropriate default from available inputSources, such as prioritizing based on the value of targetRayMode:
+  // 'screen' over 'tracked-pointer' over 'gaze'.
+  preferredInputSource = computePreferredInputSource(session.getInputSources()); // does this return null if it's gaze? should we always have a button-less gaze input for screen-base XR?
```
toji: To answer the question in the comments: It's been my understanding that in the case of a gaze-input only device (cardboard, etc.) we will populate a single inputSource that tracks the head.

Also, to clarify what you mean by "buttonless", is that implying an input source without an `XRTrackedController` or one that cannot produce `select` events? The former seems fine, but I would object to the latter.

input-explainer.md
```diff
@@ -339,6 +395,25 @@ interface XRInputSource {
   readonly attribute XRSpace? gripSpace;
 };
 
+//
+// Hit Testing
+//
+
+[SecureContext, Exposed=Window]
+interface XRHitTestSource {
```
toji: The code samples above and the interface name make it seem like this is intended to extend `XRInputSource`? Am I mis-reading it?

spatial-tracking-explainer.md
```diff
@@ -350,6 +359,7 @@ partial interface XRFrame {
   // Also listed in the main explainer.md
   XRViewerPose? getViewerPose(optional XRReferenceSpace referenceSpace);
   XRPose? getPose(XRSpace space, XRSpace relativeTo);
+  Promise<XRAnchor> requestAnchor(XRSpace relativeTo, XRRigidTransform offset);
```
toji: Oh, this is interesting. Why is it necessary to put anchor creation on the `XRFrame` object?

toji:
> "give me a hit test that will fit the size of this virtual object so I can put it there"

Is this something that's common in AR apps? For phone AR at least, what I typically see is just a drawn outline of the object to be placed around the reticle, but the hit test is still done at a single centroid. Perhaps this is more common with HoloLens/Magic Leap where there's actually an environment mesh and I just don't know it. (Gotta get myself one of those in the near future. Crossing my fingers for HoloLens 2 news soon-ish...)

spatial-tracking-explainer.md
NellWaliczek: I'd been assuming it was necessary because the XRSpace changes from frame-to-frame and to get the right spot, you'd need to know where it was on the frame the anchor is requested.

NellWaliczek:
Removing this from the PR text and putting it here to track the conversation:
> // Per our conversation, I'm contemplating the approach that an XRViewerInputSource should
// ALWAYS be present with gamepad=null.  One and only one.  Regardless of whether 
// or not there's other input sources.  It's a tad bit weird if there's a gamepad
// plugged in, but I think it's less confusing to have two XRViewerInputSource
// objects (one with gamepad and one without) to ensure that the oninputsourcechange 
// event behaves as expected should the gamepad be unplugged.
//
// The bonus to this approach is that in the event handler for select, developers can
// query the button on the gamepad that actually fired the event...
//
// Also your point about tracked controllers potentially having both hands and buttons
// can also  be applied to the XRViewerInputSource if needed.

input-explainer.md
NellWaliczek: Sounds like after our chat on the call today, that you're more comfortable with the idea of an always present viewer-based input source?  If that's not the case after you sleep on it, please reopen this!

input-explainer.md
```diff
@@ -280,6 +280,78 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+    let targetPose = null;
+    
+    let environmentHitTestResults = xrFrame.getHitTestResults(preferredInputSource.hitTestSource, xrReferenceSpace);
+    
+    let boundsHitTestResults = {};
+    foreach (key in activeDragInteraction.hitTestBounds) {
+      boundsHitTestResults[key] = xrFrame.getHitTestResults(activeDragInteraction.hitTestBounds, xrReferenceSpace);
+    }
+
+    let hitResult = scene.hitTest(new XRRay(targetRayPose.transform));
```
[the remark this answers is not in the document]
NellWaliczek: I think I've fixed this with my latest changes (though obviously not propagated them through the whole file yet)...

input-explainer.md
NellWaliczek: Honestly, I go back and forth on this one.  I'm half considering getting rid of targetRaySpace in favor of hitTestSource, tbh.... when you see the sample code, I think you may agree... lmk?

NellWaliczek:
@toji I feel like this is looking a lot better and has addressed most, if not all, the major issues we originally raised.  Obviously there's quite a few more steps to go, but this seems like it has potential.

@lincolnfrog this is the PR I mentioned to you earlier today.  Looking forward to chatting with you about it tomorrow!

input-explainer.md
```diff
@@ -280,6 +280,108 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
 
+### Environmental hit-testing
+Descriptive text goes here
+
+Basic example of using input sources to get hit test results each frame
+```js
+function onDrawFrame(timestamp, xrFrame) {
+  // Frame logic
+
+  let hitTestResults = xrFrame.getHitTestResults(preferredInputSource.hitTestSource, xrReferenceSpace);
```
lincolnfrog: I am curious about how we imagine this working on a platform that can't support synchronous hit-test requests. Do we pre-emptively do a hit-test for each input source every frame so we have the results already? Is this a performance concern? Are their other options?

input-explainer.md
```diff
@@ -280,6 +280,108 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+}
+
+function onSelectStart(event) {
+  // Ignore the event if we are already dragging
+  if (activeDragInteraction)
+    return;
+
+  // Use the input's hitTestSource to find a draggable object in the scene
+  let hitTestResult = getHitTestResult(event.frame, event.inputSource.hitTestSource);
+  if (hitTestResult && hitTestResult.target && hitTestResult.target.isDraggable) {
```
lincolnfrog: Is hitTestResult.target only valid if the hit was a virtual object from the app's scene? Or, would it be a plane if a plane were hit?

input-explainer.md
```diff
@@ -280,6 +280,108 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+
+let anchors = [];
+
+// Only end the drag when the input source that started dragging releases the select action
+function onSelectEnd(event) {
+  if (activeDragInteraction && event.inputSource == activeDragInteraction.inputSource) {
+    let hitTestResult = getHitTestResult(event.frame, activeDragInteraction.hitTestSource);
+    if (hitTestResult && !hitTestResult.isVirtual) {
+      if (MatrixMathLibrary.contains(hitTestResult.bounds, activeDragInteraction.target.footprint)) {
+        frame.requestAnchor({space:xrReferenceSpace, offset:hitTestResult, type:hitTestResult.type}).then((anchor) => {
```
lincolnfrog: Just want to make sure we think through requestAnchor being a promise - do we care about the specific timing of when the anchor resolves?

input-explainer.md
```diff
@@ -288,6 +390,8 @@ This is a partial IDL and is considered additive to the core IDL found in the ma
 //
 
 partial interface XRSession {
+  Promise<XRHitTestSource> requestHitTestSource(XRSpace space, optional XRRay ray, optional XRSpaceType spaceType);
```
lincolnfrog: Does this essentially function as a subscription in the case of a platform that can't support synchronous hit-testing (see my earlier comment for context)?

input-explainer.md
NellWaliczek: Closing the loop after our call...
The seemingly synchronous hit-test results are due to an implicit registration on behalf of the XRInputSource objects when they are created.  The underlying platform will perform hit-tests for all XRHitTestSource objects that still have references to them held and provide that data as part of the XRFrame in callbacks and events.

input-explainer.md
NellWaliczek: Yeah, this is a bit confusing because hitTestResult.target isn't part of the API.  It's something that the sample code's imaginary engine added.  I'm not sure what the right way to make that clearer is, though I am concerned about the confusion.  I'll noodle on it as I clean up the document....

input-explainer.md
NellWaliczek: Honestly, I didn't spend a ton of time digging into the Achor API shape other than as a proof of concept for how it would relate to hit test results.  I read Iker's explainer and tried to apply the info there to the pattern of this design.  I'll probably need to hold off on a specific proposal there until I've had a chance to talk to him.

input-explainer.md
NellWaliczek: Yeah, this is the "subscription" mechanism for non-InputSource objects or for offsets/type changes on the InputSource's hitTestSource

input-explainer.md
```diff
@@ -319,24 +425,88 @@ interface XRRay {
...
+
+  // TODO: Probably shouldn't add these, but I don't want to lose track of the possible approach
+  // Promise<void> updateOffsetRay(XRRay ray);
+  // Promise<void> updateType(XRSpaceType type);
+};
+
+[SecureContext, Exposed=Window]
+interface XRHitTestResult {
+  readonly attribute XRRigidTransform transform;
+  readonly attribute XRRay normal;
```
toji: This seems redundant to me. The normal should be able to be inferred by the transform's orientation (and we already have a method for constructing rays from transforms, so there's that.) Plus a normal doesn't need an origin, so if we did want to communicate the normal explicitly we'd probably want to expose it as an DOMPoint with a value of `{x,y, z, 0.0}`;

input-explainer.md
```diff
@@ -319,24 +425,88 @@ interface XRRay {
...
 
+[SecureContext, Exposed=Window]
+interface XRInputSource {
+  // TODO: Consider if we should we remove this in favor just using hitTestSource.InputSource?
+  readonly attribute XRSpace targetRaySpace; 
+
+  // hitTestSource.space = targetRaySpace
+  // hitTestSource.type = ""
+  // hitTestSource.offsetRay = FORWARD RAY
+  readonly attribute XRHitTestSource hitTestSource;
```
toji: Slightly concerned about this, simply because it implies that we're doing the hit test behind the scenes for each input source all the time. I would HOPE that doesn't have a high implicit cost, but I'm not sure we can guarantee it. Also, if you encountered a platform that had some sort of silly limit like "We can only process 5 hit tests per frame" you may end up inadvertently eating up a large chunk of that simply because you have 2-3 input sources. Also, as pointed out on our call, the developer will have to request a new hit test source anyway if we guess their intentions wrong. (IE: an off-center hit ray for screen reticles)

As a result, I suggest we simplify the API surface by just letting the developer request their own hit test sources for input spaces when they want them.

input-explainer.md
NellWaliczek: In the other repo, I decided to give it a try and removed targetRaySpace.  Take a look and let me know what you think over there in #20

input-explainer.md
NellWaliczek: Hopefully this is much clearer in #20 If not, lmk specific points of confusion on that PR and I'll try to clear it up.

input-explainer.md
```diff
@@ -280,6 +280,108 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+    return;
+
+  var realHitTestResults = frame.getHitTestResults(hitTestSource, xrReferenceSpace);
+
+  let targetRayPose = frame.getPose(hitTestSource, xrReferenceSpace);
+  if (targetRayPose) {
+    var virtualHitTestResult = scene.hitTest(new XRRay(targetRayPose.transform));
+  }
+
+  if (virtualHitTestResult && realHitTestResults[0]) {
```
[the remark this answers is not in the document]
NellWaliczek: Let me know if you still feel that way when reading #20?

input-explainer.md
NellWaliczek: I cleaned this up a bunch in PR #20.  Let me know over there if it's still confusing?

input-explainer.md
NellWaliczek: The reason I went with this approach was to address the issue where screen-based inputs will have a delay before being able to report hit test results.  This way, as soon as the screen-based input source is present, it's able to report hit test results.
It's worth noting that an XRHitTestSource isn't guaranteed to return a hit point, so User Agents that have the numeric limitation you mentioned can use a heuristic to determine which ones are important (for example: only the ones the developer is actively querying?).  That seems like a fairly theoretical limitation.  @lincolnfrog  didn't seem super concerned about it earlier today, but you're right that it's worth talking to Alex, Artem, and others to get a double check of their platforms.
@lincolnfrog can you chime in here?

input-explainer.md
```diff
@@ -319,24 +425,88 @@ interface XRRay {
...
 };
 
 [SecureContext, Exposed=Window]
-interface XRInputSource {
-  readonly attribute XRHandedness handedness;
-  readonly attribute XRTargetRayMode targetRayMode;
-  readonly attribute XRSpace targetRaySpace;
-  readonly attribute XRSpace? gripSpace;
+interface XRViewerInputSource : XRInputSource {
+  readonly attribute Gamepad? gamepad;
```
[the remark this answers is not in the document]
NellWaliczek: That's not the purpose of this... at least with Windows Mixed Reality, we fired a "select" event for gamepad buttons, so this would allow you to know which gamepad to examine the buttons of.  That said, I'm not super strongly attached to that approach.  It was more of a developer convenience.

input-explainer.md
```diff
@@ -319,24 +425,88 @@ interface XRRay {
...
+  "",
+  "vertical-plane",
+  "horizontal-plane",
+  // in the future maybe "bounded-volume" ?
+};
+
+dictionary XRHitTestSourceOptions {
+  required XRSpace space;
+  attribute XRRay offsetRay = new XRRay();
+  XRSpaceType type = "";
```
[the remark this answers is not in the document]
NellWaliczek: I'm happy to remove this from the initial PR, though it does make the grab-and-drag sample really nice.  I'd be curious to hear @lincolnfrog 's opinion...

input-explainer.md
NellWaliczek: I decided to change this to have XRInputSource inherit from XRHitTestSource.  Take a look at #20 and lmk what you think over there

input-explainer.md
NellWaliczek: That's a fair point.  I'll remove it.

input-explainer.md
```diff
@@ -319,24 +425,88 @@ interface XRRay {
   readonly attribute Float32Array matrix;
 };
 
+[SecureContext, Exposed=Window]
+interface XRInputSource {
+  // TODO: Consider if we should we remove this in favor just using hitTestSource.InputSource?
```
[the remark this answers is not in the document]
NellWaliczek: Ah, I have a bug in #20 that I'll fix right now, but no.. you can still draw the ray.

input-explainer.md
```diff
@@ -280,6 +280,108 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+function getHitTestResult(frame, hitTestSource) {
+  let hitTestResult = null;
+
+  let viewerPose = frame.getViewerPose(xrReferenceSpace);
+  if (!viewerPose || !hitTestSource)
+    return;
+
+  var realHitTestResults = frame.getHitTestResults(hitTestSource, xrReferenceSpace);
+
+  let targetRayPose = frame.getPose(hitTestSource, xrReferenceSpace);
```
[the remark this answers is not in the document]
NellWaliczek: D'oh.  I'll fix this.

input-explainer.md
```diff
@@ -280,6 +280,108 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+  if (targetRayPose) {
+    var virtualHitTestResult = scene.hitTest(new XRRay(targetRayPose.transform));
```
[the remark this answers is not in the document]
NellWaliczek: Take a look at how I've structure #20 and lmk over there if you think this is still an issue

input-explainer.md
NellWaliczek: Fixed in #20

input-explainer.md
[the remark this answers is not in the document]
NellWaliczek: Fixed in #20

input-explainer.md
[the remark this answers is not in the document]
NellWaliczek: Fixed in #20

NellWaliczek:
Ok, now that I've got the other PRs up in editor-collab, I'm going to close this one.  Until the final thing is merged, I'm going to leave the branch active.

PR #20, opened 2019-01-09
Hit testing

NellWaliczek:
So, this maybe got a little bit out of hand.  ^_^;

The current state of things is:
* The IDL is what I'm currently proposing
* All sample code is updated to the IDL, but hasn't been proofread
* The document has been restructured for a clearer flow
* Some of the prose has been updated, but mostly not

Tomorrow I'll wrap up the remaining prose and then have you do a full review.  If you want to just look at the IDL now, feel free since a made a few minor tweaks since you looked at it :)

That said, I'm not sure anyone is going to be able to review this diff meaningfully given how many changes to the document structure I made as part of cleaning up the samples.  Do you think the benefit to clarity is going to be worth the hassle to folks to have to review the whole document from scratch?


NellWaliczek:
Oh, one other thing... when I post this to the real WebXR repo, I'm thinking to include an example of how it *might* work with anchors in the PR description so that people can see the connection before the actual PR is merged.  What do you think?

explainer.md
```diff
@@ -223,14 +223,18 @@ function onDrawFrame(timestamp, xrFrame) {
   if (xrSession) {
     let glLayer = xrSession.renderState.baseLayer;
     let pose = xrFrame.getViewerPose(xrReferenceSpace);
-    gl.bindFramebuffer(gl.FRAMEBUFFER, glLayer.framebuffer);
+    if (pose) {
+      // Run imaginary 3D engine's simulation to step forward physics, animations, etc.
+      scene.updateScene(timestamp, xrFrame);
```
toji: Seems likely the engine wants to tick physics, etc forward even in cases without a valid pose? Move outside the `if`?

input-explainer.md
```diff
@@ -1,284 +1,407 @@
...
 
-### Enumerating input sources
-
-Calling the `getInputSources()` function on an `XRSession` will return a list of all `XRInputSource`s that the user agent considers active. An `XRInputSource` may represent a tracked controller, inputs built into the headset itself, or more ephemeral input mechanism like tracking of hand gestures. When input sources are added to or removed from the list of available input sources the `inputsourceschange` event will be fired on the `XRSession` object to indicate that any cached copies of the list should be refreshed.
+An `XRInputSource` may represent a tracked controller, inputs built into the headset itself, or more ephemeral input mechanism like tracking of hand gestures.
```
toji: May be good to mention screen space input in this list, since it's a concrete scenario most devs will encounter.

toji: (To be clear, I know this is old text relocated. Just adding a note to make sure I don't forget the request.)

input-explainer.md
```diff
@@ -1,284 +1,407 @@
...
-### Enumerating input sources
-
-Calling the `getInputSources()` function on an `XRSession` will return a list of all `XRInputSource`s that the user agent considers active. An `XRInputSource` may represent a tracked controller, inputs built into the headset itself, or more ephemeral input mechanism like tracking of hand gestures. When input sources are added to or removed from the list of available input sources the `inputsourceschange` event will be fired on the `XRSession` object to indicate that any cached copies of the list should be refreshed.
+An `XRInputSource` may represent a tracked controller, inputs built into the headset itself, or more ephemeral input mechanism like tracking of hand gestures. 
+The properties of an XRInputSource object are immutable. If a device can be manipulated in such a way that these properties can change, the `XRInputSource` will be removed and recreated.
 
-```js
-// Get the current list of input sources.
-let xrInputSources = xrSession.getInputSources();
+The `targetRaySpace` will never be `null`. It's value will differ based on the type of input source that produces it, which is represented by the `targetRayMode` attribute.
```
toji: This paragraph feels out of place.

toji: Oh, I see. It's the lead-in paragraph to the list of modes below. That wasn't clear to me on first readthrough. Maybe make `## Target Rays` a section header with the types as subsections?

input-explainer.md
```diff
@@ -1,284 +1,407 @@
...
+  // Scene update logic ...
+
+  let hitTestResults = xrFrame.getHitTestResults(preferredInputSource, xrReferenceSpace);
+
+  // Other scene update logic ...
 }
 ```
 
-### Rendering input sources
+### Real hit test
```
toji: I think you got these headers swizzled

input-explainer.md
```diff
@@ -1,284 +1,407 @@
...
-  // Use the input source target ray to find a draggable object in the scene
-  let hitResult = scene.hitTest(new XRRay(targetRayPose.transform));
-  if (hitResult && hitResult.draggable) {
-    // Use the targetRayPose position to drag the intersected object.
+  // Use the input's hitTestSource to find a draggable object in the scene
+  let hitTestResult = getHitCombinedHitTestResult(event.frame, inputSource);
+  
+  // The virtualTarget object isn't part of the WebXR API.  It is
+  // something set by the imaginary 3D engine in this example
+  let target = hitTestResult.virtualTarget;
```
toji: One of my concerns here is that even though the comment calls it out as not part of the API, this isn't ever shown as being set in the code above which produces the hit test results. I would also prefer an approach that doesn't appear to patch new attributes onto native objects just to avoid confusion.

input-explainer.md
lincolnfrog: +1

toji:
Added some comments, and wanted to follow up with the broad observation that it seems like ~80% of this PR is not actually hit test related. I like the general direction of the explainer updates, and many of them feel like they make sense outside the context of the hit test change, so I'd really like to see those land separately so that the hit test PR is much more focused on the hit test functionality. (I'm assuming this is a "Nell already knows this and plans on splitting after we can look at the changes in-context" thing, but it's worth stating anyway. 😄)

input-explainer.md
```diff
@@ -1,284 +1,407 @@
...
 
-The `gripSpace` represents a space where if the user was holding a straight rod in their hand it would be aligned with the negative Z axis (forward) and the origin rests at their palm. This enables developers to properly render a virtual object held in the user's hand. For example, a sword would be positioned so that the blade points directly down the negative Z axis and the center of the handle is at the origin.
+When the canvas receives a `pointerdown` event an `XRInputSource` is created with a `targetRayMode` of `'screen'` and added to the array returned by `getInputSources()`. A `selectstart` event is then fired on the session with the new `XRInputSource`. The `XRInputSource`'s target ray should be updated with every `pointermove` event the canvas receives until a `pointerup` event is received. A `selectend` event is then fired on the session and the `XRInputSource` is removed from the array returned by `getInputSources()`. When the canvas receives a `click` event a `select` event is fired on the session with the appropriate `XRInputSource`.
```
lincolnfrog: Do we also need to create an XRHitTestSource automatically for this as well? If we don't, you still need to do an async request to generate one from the XRInputSource, which will delay using the touch event by a frame. I might be missing something...

input-explainer.md
```diff
@@ -1,284 +1,407 @@
...
+  xrSession.addEventListener('inputsourceschange', onInputSourcesChange);
+  onInputSourcesChange();
 
-  // Render a visualization of the input source if appropriate.
-  // (See next section for details).
-  renderInputSource(xrFrame, inputSource);
+  // More session initialization logic ...
+}
+
+function onInputSourcesChange() {
```
lincolnfrog: Should this just take in the list of input sources to make it simpler / more direct?

input-explainer.md
```diff
@@ -1,284 +1,407 @@
...
+  if (hitTestSourcePose) {
+    // Invoke the example 3D engine to compute the virtual hit test
+    var virtualHitTestResult = scene.hitTest(new XRRay(hitTestSourcePose.transform));
+  }
+
+  // Other scene update logic ...
+}
+```
+
+### Combined hit testing
```
lincolnfrog: I think having this example is really great

input-explainer.md
toji: I raised the same concern about input latency on an editors call. It's something that I get more concerned about the more I think about it because it's the type of thing that imposes an arbitrary "web tax" on experiences. Nell and I talked about the possibility of making this opt-in/out for experiences that want it.

input-explainer.md
```diff
@@ -1,284 +1,407 @@
...
-While the primary motivation of this input model is a compatible "target and click" interface, more complex interactions such as grabbing and dragging with input sources can also be achieved using only the `select` events.
+### Custom hit testing
+Expand on this sample
 
 ```js
-// Stores details of an active drag interaction, is any
-let activeDragInteraction = null;
+let xrHitTestSource = null;
+  function createNewHitTestSource() {
+  let hitTestOptions = { space:preferredInputSource.space, resultTypes:["horizontal-plane"] };
```
lincolnfrog: So you are limited to the ray of the input source? Is this too limiting (vs. allowing an arbitrary ray)?

input-explainer.md
```diff
@@ -319,26 +446,56 @@ interface XRRay {
...
+
+enum XRHitTestResultType {
+  "point",
+  "vertical-plane",
+  "horizontal-plane",
+};
+
+dictionary XRHitTestSourceOptions {
+  required XRSpace space;
+  XRRay offsetRay = new XRRay();
```
lincolnfrog: Ah, good, I didn't see this. Will update my comment above.

input-explainer.md
```diff
@@ -319,26 +446,56 @@ interface XRRay {
...
+  Array<XRHitTestResultType> resultTypes = "point"; 
+};
+
+// TODO: Consider if this should inherit from XRSpace instead?
```
lincolnfrog: As we discussed, I prefer this "has-a" setup, it makes more sense.

input-explainer.md
lincolnfrog: Ah, found below the full list of options. Maybe use the offsetRay option here to avoid this confusion for others?

input-explainer.md
lincolnfrog: +1

toji:
Following up from another call topic: I feel like we do probably need a more lightweight "one-shot" hit test mechanism with an arbitrary ray regardless of how things like screen input is handled.

Goofy example: I have an AR game where you throw darts at paint-filled balloons. When they explode the balloons spray paint decals all over your environment. This can technically work by creating a hit test source for each directional splatter, but that starts to feel an awful lot like fighting with the API to do something that should otherwise be simple.

(Aside: my kids would play the aforementioned app for HOURS)

I wonder if we could solve the issue of clicking and dragging by allowing a one-shot hit test to be "upgraded" into a hit test source once you determine that it's hit something interesting.

NellWaliczek:
I'll take a stab at addressing the comments and fixing up the prose today and tomorrow.  

*sigh* yes, I was sort of afraid you'd say that about splitting the PR.  Well, step one is to clean up what I've got and then we can look at the whole thing and figure out if there's a good way to remove the real-world hit-testing portion while leaving the virtual hit-testing portion.  Otherwise, the sample code is going to need a lot of rewriting :-/

As for your goofy example.... tbh, that's not something I'd think we'd be aiming to support with the hit-testing API.  That's something a full mesh/point-cloud solution should be used for.  And I'm hesitant to contort the hit-test functionality to support it.

toji:
After mulling on it for the past day I'm feeling more confident that my above proposal could work out. Sketching out the related IDL, it might look something like this:

```webidl
// Simplifying for the moment to focus on relevant bits
interface XRHitTestSource {
  readonly attribute XRSpace space;
  readonly attribute XRRay offsetRay;
};

interface XRHitTestResult {
  readonly attribute XRRigidTransform transform;
  // If the hit test was produced from an XRHitTestSource it simply returns that, otherwise
  // we return a new one immediately using the same space and offset ray that produced
  // this hit test.
  XRHitTestSource getHitTestSource();
};

dictionary XRHitTestOptions { // Note the minor name change.
  required XRSpace space;
  XRRay offsetRay = new XRRay();
}

partial interface XRSession {
  Promise<XRHitTestSource> requestHitTestSource(XRHitTestOptions options);
};

partial interface XRFrame {
  FrozenArray<XRHitTest> getHitTestResults(XRHitTestSource hitTestSource, optional XRSpace relativeTo);
  Promise<FrozenArray<XRHitTest>> getAsyncHitTestResults(XRHitTestOptions options, optional XRSpace relativeTo);
};
```

toji:
Yay for synchronized comment posting!

> that's not something I'd think we'd be aiming to support with the hit-testing API.

This is slightly surprising to me, as I'm not sure in that case where you envision the line being drawn for the types of things our hit testing API should support?

NellWaliczek:
Oh geez... I totally misread your proposal.  Yes, that would get you the hit-test result for the frame in which you requested it.  Sorry!

NellWaliczek:
Though hmmmm.... whether you have a `getHitTestSource()` or a `hitTestSource` object, implicitly that's keeping the hit-test capability alive and would be a performance hit... perhaps just the space/ray attributes should be there?

NellWaliczek:
The latest push is still far from complete... but I added an idea I thought of to address the performance concerns while still being responsive for screen-based input sources.  The rough idea in the very last commit, but I haven't finished updating sample code. Basically, we automatically create the XRHitTestSources and pass them into the inputsourceschanged event and if they aren't saved, well so what?  Then no performance hit and if they change their mind, the dev can *always* create a new one with the XRInputSource itself.  The syntax in the event isn't great yet, but I can work on that.

Also!  I removed the requirement to automatically create an viewer-based input source and put a viewerSpace on session (that was in an earlier commit)

So yeah, the one real catch I see here is that the inputsourcechanged event will need to fire once at the beginning of the session.  Probably after the session creation callback is done being invoked

WDYT?

input-explainer.md
NellWaliczek: This should be fixed with the latest prose.

input-explainer.md
NellWaliczek: *sigh* ok.  I'll tackle this shortly

input-explainer.md
NellWaliczek: Yes, a XRHitTestSource will be created for screen-based input automatically.  That said, I'm pretty sure my latest proposal fixes this issue by decoupling the XRInputSource type from the XRHitTestSource type.  Take a look and let me know if you still have questions or concerns.

input-explainer.md
NellWaliczek: Yay!

input-explainer.md
NellWaliczek: This comment was actually the inspiration for my solution to the performance problem!  Thanks!

input-explainer.md
NellWaliczek: Hopefully this is resolved adequately.  Please reopen if not!

input-explainer.md
NellWaliczek: Me too.  It also made @toji 's desire for one-off hit testing much easier.

NellWaliczek:
Fun fun!  I just pushed another few commits that hopefully addressed most of the design issues.  Obvi I'm still updating prose, but it's probably worth another look.  I'd suggest this diff to see what's changed since your last looking at it:
https://github.com/immersive-web/editor-collab/pull/20/files/4119e7354025b3f5f852cd60151eaa2893fe87bc..849e239778563fcfe9c80275e509dcc4bffad698

input-explainer.md
```diff
@@ -304,6 +494,9 @@ partial interface XRFrame {
   // Also listed in the spatial-tracking-explainer.md
   XRViewerPose? getViewerPose(optional XRReferenceSpace referenceSpace);
   XRPose? getPose(XRSpace space, XRSpace relativeTo);
+
+  FrozenArray<XRHitTest> getHitTestResults(XRHitTestSource hitTestSource, optional XRSpace relativeTo);
```
lincolnfrog: I don't see a definition for XRHitTest anywhere. It seems to me that we have a slightly wrong structure here. What it seems like it should be is XRHitTest is effectively a tuple of (XRFrame, FrozenArray<XRHitResult>), so these getHitTestResults/requestAsyncHitTestResults should return a Promise<XRHitTest>? I am also not convinced that these names are super clear - especially XRHitTest, but we need to figure out what it contains first.

input-explainer.md
```diff
@@ -1,284 +1,470 @@
...
-// Update the list of input sources if it ever changes.
-xrSession.addEventListener('inputsourceschange', (ev) => {
+  xrSession.addEventListener('inputsourceschange', onInputSourcesChange);
+
+  // More session initialization logic ...
+}
+
+let xrInputSources = null;
+function onInputSourcesChange(event) {
   xrInputSources = xrSession.getInputSources();
```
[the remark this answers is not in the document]
NellWaliczek: Ah, I hadn't realized that's what you were asking.  I updated the sample to grab the session off the event.

input-explainer.md
```diff
@@ -1,284 +1,470 @@
...
 
+> TODO fit this in somewhere
+
+In some cases tracked input sources cannot accurately track their position in space, and provides an estimated position based on the sensor data available to it. This is the case, for example, for the Daydream and GearVR 3DoF controllers, which use an arm model to approximate controller position based on rotation. In these cases the `emulatedPosition` attribute of the `XRInputPose` should be set to `true` to indicate that the translation components of the pose matrices may not be accurate.
+
+### Hit test sources
+Prose that describes the asynchronicity issues.  Also explain that all getHitTestSource calls will return null for non-AR sessions.
+
+#### Automatic creation
+Explain why this is useful to ensure that hit test results are always available without a delay.  Also explain that because of the performance cost, developers may not choose to keep the automatically created hit test sources.  Also explain that `event.getHitTestSource()` will return null unless the parameter is in the `event.added` array.
```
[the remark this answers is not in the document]
NellWaliczek: Yeah, this prose was just scribbled notes to myself of stuff not to forget ^_^

input-explainer.md
```diff
@@ -1,284 +1,470 @@
...
-  * Clicking a touchpad
-  * Tapping a button
-  * Making a hand gesture
-  * Speaking a command
-  * Clicking or touching a canvas.
+```js
+let viewerHitTestSource = null;
 
-Three events are fired on the `XRSession` related to these primary actions: `selectstart`, `selectend`, and `select`.
+let hitTestOptions = { space:xrSession.viewerSpace, offsetRay:new XRRay({ origin: {y: 0.5} }) };
```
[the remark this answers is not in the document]
NellWaliczek: What do you mean by this?
> adding a ray as an offset to another ray

input-explainer.md
NellWaliczek: Ah, it was just a typo.  That was supposed to be `XRHitTestResult` in both places.  Fixed now.

Not sure what's leading you to think that the `getHitTestResults()` call should be promise based?  The data should already be part of the frame, yes?  

As for `requestHitTestResults()`, that already is returning a promise, and when the promise is completed the data will be relative to the `XRFrame` on which the function was called.  I know that part isn't in the prose yet.. sorry for the confusion.

NellWaliczek: I think this is cleared up in the current version.  If there's still confusion, please reactivate!

NellWaliczek:
As part of preparation to split into multiple PRs, I've added a separate explainer for the hit-testing content.  I also squashed a bunch of minor commits, rebased, and targeted master now that the pose changes are in.

Still a bunch more text changes to be made, but this is feeling like solid progress.

environment-awareness-explainer.md
```diff
@@ -0,0 +1,321 @@
...
+
+#### Requesting a hit test source
+To create an `XRHitTestSource` developers call the `XRSession.requestHitTestSource()` function.  This function accepts an `XRHitTestOptionsInit` dictionary with the following values:
+* The `space` is required and is the `XRSpace` to be tracked by the hit test source.
+* The `offsetRay` is optional and is the `XRRay` from which the hit test should be performed. This ray is defined relative to the `space`. If an `offsetRay` is not provided, the default is an `XRRay` with a forward direction and an origin coincident with the `space` origin.
+
+>**QUESTION** Should we leave out the `featureType` for the first PR?  If so, I'll need to change the grab-and-drag sample.
```
toji: I would be more comfortable leaving it out of the first PR, yes. It's one of the aspects of this proposal that has the potential to be more contentious and privacy sensitive.

environment-awareness-explainer.md
```diff
@@ -0,0 +1,321 @@
...
+
+>**TODO** figure out how to know hand-held vs. head-worn and show this sample based on that
```
toji: Do you have an example of a case where this matters? From my understanding it seems like you probably want "the best pointing mechanism you support" in which case you defer to whatever the inputSources gives you, or you explicitly want the user's head, in which case you use the viewer space?

environment-awareness-explainer.md
```diff
@@ -0,0 +1,321 @@
...
+
+In this example, an `XRHitTestSource` is created slightly above the center of the `viewerSpace`. This is because the developer has chosen to leave space for buttons along the bottom of the hand-held AR device screen while still giving the perception of a centered cursor (see [Rendering cursors and highlights](#Rendering-cursors-and-highlights)
```
toji: 👍 Nice example choice, very real-world.

environment-awareness-explainer.md
```diff
@@ -0,0 +1,321 @@
...
+
+```js
+let viewerHitTestSource = null;
+
+let hitTestOptions = { space:xrSession.viewerSpace, offsetRay:new XRRay({ origin: {y: 0.5} }) };
```
toji: This isn't how the `XRay` constructor is defined. It would be `new XRRay({y: 0.5})` with the current constructor shape. (We can change that, of course, if you think it's necessary.)

environment-awareness-explainer.md
```diff
@@ -0,0 +1,321 @@
...
+
+To mitigate this difficulty, for each new `XRInputSource` user agents must automatically generate an `XRHitTestSource` to be retrievable by the `XRInputSourcesChangeEvent.getHitTestSource()` function.  When passed an `XRInputSource` from the `XRInputSourcesChangeEvent.added` attribute, the pre-generated `XRHitTestSource` will be returned.  Any other parameter will return throw a `DOMException`. It's worth noting that, as mentioned above, each active `XRHitTestSource` may have a negative performance impact.  Therefore instead of an implicit binding of `XRHitTestSource` objects to the lifetime of `XRInputSource` objects, developers must explicitly choose to retain the `XRHitTestSource` if they wish to use it. Otherwise, at the end of the `inputsourceschange` callbacks the pre-created hit test source will disappear.
```
toji: This definitely feels like a better pattern to me than having a hit test source perpetually on every input source, but I still do worry that it implies added latency for things like touch events. My worry would be that in order to resolve that implementations may implicitly delay wiring up the hit test source until it's requested and simply return `null`'s in the meantime, at which point the same several-frame-long delay is still effectively incurred.

environment-awareness-explainer.md
```diff
@@ -0,0 +1,321 @@
...
+  // Also listed in the input-explainer.md
+  attribute EventHandler oninputsourceschange;
+};
+
+//
+// Frame
+//
+
+partial interface XRFrame {
+  FrozenArray<XRHitTestResult> getHitTestResults(XRHitTestSource hitTestSource, optional XRSpace relativeTo);
```
toji: `getHitTestSourceResults()`, maybe?

environment-awareness-explainer.md
```diff
@@ -0,0 +1,321 @@
...
+}
+```
+
+On occasion, developers may want hit test results for the current frame even if they have not already created an `XRHitTestSource` to subscribe to the results. For example, when a virtual object needs to bounce off a real-world surface, a single hit-test result can be requested.  The result will be delivered asynchronously, though it will be accurate for the frame on which it was requested.
+
+```js
+function updateScene(timestamp, xrFrame) {
+  // Scene update logic ...
+
+  xrFrame.requestHitTestResults(hitTestSources[preferredInputSource], xrReferenceSpace).then((results) => {
```
toji: In the IDL below you have this as `requestAsyncHitTestResults()`

environment-awareness-explainer.md
```diff
@@ -0,0 +1,321 @@
...
+    // do something with results
+  });
+
+  // Other scene update logic ...
+}
+```
+
+### Combining virtual and real-world hit testing
+A key component to creating realistic presence in XR experiences, relies on the ability to know if a hit test intersects virtual or real-world geometry first. For example, a virtual button should be "clicked" if the user selects it, but not if there's a physical object in the way. There are a handful to different techniques which can be used to determine a combined hit test result that take into account a user's previous actions.  For example, a developer may choose to weight hit test results differently if a user is already interacting with a particular object.
```
toji: Note from offline conversation: Developers probably only want to consider real world geometry first if the virtual geometry is visibly occluded by it, which we don't have a mechanism for yet (and is highly unlikely in V1, I would think.) ☹️

environment-awareness-explainer.md
```diff
@@ -0,0 +1,321 @@
...
+
+  if (inputSource) {
+    let inputSourcePose = frame.getPose(inputSource.source, xrReferenceSpace);
+    if (inputSourcePose) {
+      var virtualHitTestResult = scene.virtualHitTest(new XRRay(inputSource.transform));
+    }
+  }
+
+  // Explain that this is a grossly simplified way of determining which hit test result to use
+  let isVirtualCloser = false;
```
toji: Because of the occlusion problem mentioned above this should just assume that if a virtual hit occurred it is the one that gets used for now. Text preceding the example should be updated to account for that as well.

environment-awareness-explainer.md
```diff
@@ -0,0 +1,321 @@
...
+    if (virtualTarget && virtualTarget.isDraggable) {
+      activeDragInteraction = {
+        inputSource: inputSource,
+        target: virtualTarget,
+        initialTargetTransform: virtualTarget.transform,
+        initialHitTestResult: combinedHitTestResult["result"]
+      };
+
+      // Change the hit-testing to look for a horizontal plane
+      let hitTestOptions = { space:preferredInputSource, featureType:"horizontal-plane" };
```
toji: As was mentioned earlier, I'd feel more comfortable if we deferred the `featureType` portion of this.

environment-awareness-explainer.md
```diff
@@ -0,0 +1,321 @@
...
+  "vertical-plane",
+  "horizontal-plane",
+};
+
+//
+// Events
+//
+
+partial interface XRInputSourceChangeEvent {
+  XRHitTestSource getHitTestSource(XRInputSource inputSource);
```
toji: Is the implication here that I could get a hit test source for any input source in the current `getInputSources` array, even if it's not new?

NellWaliczek: `added` is defined in the input-explainer.md

NellWaliczek:
Ok, now that I've got the other PRs up in editor-collab, I'm going to close this one.  Until the final thing is merged, I'm going to leave the branch active.

PR #22, opened 2019-01-18
Add XRSession.viewerSpace and tidy XRSpace explanations

NellWaliczek:
This PR addresses two related things I noticed now that the `getPose()` changes have landed
1. We were missing a section that focused on the descriptions of `XRSpace` and `XRPose` themselves.  I added this to spatial-tracking-explainer.md and moved the `XRRigidTransform` and `XRRay` types to be together in that document's idl section
2. Having an `XRSpace` representing the viewer would be really handy for a few reasons. First, it would allow pose data to be queried relative to the viewer, such as "where are the hands relative to the headset?", without having to do extra matrix math.  Second, it's easier to describe the consistent `targetRaySpace` for 'gaze' based input sources.  Third, it becomes one of the `XRSpaces` from which hit testing can be done (see other PR).

### DIFF
--- a/spatial-tracking-explainer.md
+++ b/spatial-tracking-explainer.md
@@ -1,11 +1,11 @@
 # WebXR Device API - Spatial Tracking
-This document is a subsection of the main WebXR Device API explainer document which can be found [here](explainer.md).  The main explainer contains all the information you could possibly want to know about setting up a WebXR session, the render loop, and more.  In contrast, this document covers the technology and APIs for tracking users' movement for a stable, comfortable, and predictable experience that works on the widest range of XR hardware.
+This document explains the technology and portion of the WebXR APIs used to track users' movement for a stable, comfortable, and predictable experience that works on the widest range of XR hardware. For context, it may be helpful to have first read about [WebXR Session Establishment](explainer.md), and [Input Mechanisms](input-explainer.md).
 
 ## Introduction
 A big differentiating aspect of XR, as opposed to standard 3D rendering, is that users control the view of the experience via their body motion.  To make this possible, XR hardware needs to be capable of tracking the user's motion in 3D space.  Within the XR ecosystem there is a wide range of hardware form factors and capabilities which have historically only been available to developers through device-specific SDKs and app platforms. To ship software in a specific app store, developers optimize their experiences for specific VR hardware (HTC Vive, GearVR, Mirage Solo, etc) or AR hardware (HoloLens, ARKit, ARCore, etc).  WebXR  development is fundamentally different in that regard; the Web gives developers broader reach, with the consequence that they no longer have predictability about the capability of the hardware their experiences will be running on.
 
 ## Reference Spaces
-The wide range of hardware form factors makes it impractical and unscalable to expect developers to reason directly about the tracking technology their experience will be running on.  Instead, the WebXR Device API is designed to have developers think upfront about the mobility needs of the experience they are building which is communicated to the User Agent by explicitly requesting an appropriate `XRReferenceSpace`.  The `XRReferenceSpace` object acts as a substrate for the XR experience being built by establishing guarantees about supported motion and providing a space in which developers can retrieve `XRViewerPose` and it's view matrices.  The critical aspect to note is that the User Agent (or underlying platform) is responsible for providing consistently behaved lower-capability `XRReferenceSpace` objects even when running on a higher-capability tracking system. 
+The wide range of hardware form factors makes it impractical and unscalable to expect developers to reason directly about the tracking technology their experience will be running on.  Instead, the WebXR Device API is designed to have developers think upfront about the mobility needs of the experience they are building which is communicated to the User Agent by explicitly requesting an appropriate `XRReferenceSpace`.  The `XRReferenceSpace` object acts as a substrate for the XR experience being built by establishing guarantees about supported motion and providing a space in which developers can retrieve `XRViewerPose` and its view matrices.  The critical aspect to note is that the User Agent (or underlying platform) is responsible for providing consistently behaved lower-capability `XRReferenceSpace` objects even when running on a higher-capability tracking system. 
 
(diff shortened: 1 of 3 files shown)


spatial-tracking-explainer.md
```diff
@@ -1,16 +1,32 @@
...
+The wide range of hardware form factors makes it impractical and unscalable to expect developers to reason directly about the tracking technology their experience will be running on.  Instead, the WebXR Device API is designed to have developers think upfront about the mobility needs of the experience they are building which is communicated to the User Agent by explicitly requesting an appropriate `XRReferenceSpace`.  The `XRReferenceSpace` object acts as a substrate for the XR experience being built by establishing guarantees about supported motion and providing a space in which developers can retrieve `XRViewerPose` and its view matrices.  The critical aspect to note is that the User Agent (or underlying platform) is responsible for providing consistently behaved lower-capability `XRReferenceSpace` objects even when running on a higher-capability tracking system.
```
toji: Side note: I just love how GitHub highlights an entire paragraph as "changed" when you remove a single erroneous character. Makes it super fun to find the change. 👍 🙄

spatial-tracking-explainer.md
```diff
@@ -1,16 +1,32 @@
...
 
+### Spaces, Poses, and transforms
+XR platforms provide a variety of environment-related features such as _hit testing_, _anchors_, _spatial meshes_, _point clouds_, and _markers_. In WebXR, these features, such as `XRReferenceSpace`s, are built on top of `XRSpace`.  An `XRSpace` is an identifiable and trackable point in 3D space that tracking systems use sensor data to continuously update and adjust. On a frame-by-frame basis, developers query the locatability of an `XRSpace` via the `XRFrame.getPose()` function.
```
toji: This opening paragraph reads a little awkwardly to me, and I feel like the structure might make it easy to mis-read. Also, not sure it's accurate to say that a space is a "trackable point", as that's not really true for things like, say, the unbounded reference space, right?

Taking a stab at this, which you're free to use as much or as little as you'd like:

> One of the core features of any XR platform is it's ability to track spatial relationships. Tracking the location of the viewer is perhaps the simplest example, but many other XR platform features, such as _hit testing_ or _anchors_, are rooted in understanding the space the XR system is operating in. In WebXR any feature that tracks spatial relationships is built on top of the `XRSpace` interface. Each `XRSpace` represents something being tracked by the XR system, such as an `XRReferenceSpace`, and on a frame-by-frame basis, developers can query the location of any `XRSpace` relative to another `XRSpace` via the `XRFrame.getPose()` function.

🤷‍♀️ (Took out the reference to _spatial meshes_, _point clouds_, and _markers_ simply because I think readers might come away with the impression that WebXR supports those.)

spatial-tracking-explainer.md
```diff
@@ -1,16 +1,32 @@
...
+
+The `getPose()` function takes two parameters, `space` which is the `XRSpace` to locate and the `relativeTo` parameter which defines the coordinate system in which the resulting `XRPose` should be returned. Developers should always check the result from `getPose()` as it will be `null` on frames in which `space`'s location can not be determined within `relativeTo`. This may happen due to overall tracking loss, `space` or `relativeTo` not being locatable, or for other reasons. While the `relativeTo` parameter is an `XRSpace`, developers will often choose to supply a `XRReferenceSpace` as the `relativeTo` parameter so that coordinates will be consistent with those used for rendering, as explained in this document.
```
toji: > as explained in this document.

Is that intended to be a link to the main explainer, or to mean the spatial tracking explainer? I don't think the spatial tracking explainer really explains rendering much?

spatial-tracking-explainer.md
```diff
@@ -91,12 +107,6 @@ function onSessionStarted(session) {
 
 There is no mechanism for getting a floor-relative _unbounded_ reference space. This is because the user may move through a variety of elevations (via stairs, hills, etc), making identification of a single floor plane impossible.
 
-> **An aside regarding XRSpace**
```
toji: This always felt like a bit of a weird section to me, so I'm glad we're cleaning it up.

spatial-tracking-explainer.md
NellWaliczek: Thanks!  Yeah, I think by the time I got to that one my brain was mush.  Words are so hard. ;)

toji: That would make sense, yeah.

spatial-tracking-explainer.md
```diff
@@ -187,29 +171,10 @@ function onSessionStarted(session) {
...
 
-This type of reference space is requested with a type of `identity` and returns a basic `XRReferenceSpace`. `XRViewerPose` objects retrieved with this reference space will have a transform that is equal to the reference space's `originOffset`. The pose will only contain a single `XRView`, which must also have a transform equal to the `originOffset`.
+This type of reference space is requested with a type of `identity` and returns a basic `XRReferenceSpace`. Calls to `getPose()` with this reference space as either the `space` or `relativeTo` parameter will return null.  `XRViewerPose` objects retrieved with this reference space will have a transform that is equal to the reference space's `originOffset`. The pose will only contain a single `XRView`, which must also have a transform equal to the `originOffset`. For more information on `originOffset` see the [Application-supplied transforms section](#Application-supplied-transforms).
```
toji: :+1:  Make sure to note that this now addresses WebXR issue 478 when you move it over to that repo.

toji: Oh, and also I think the bit about being restricted to a single view is wrong now if we're allowing it in immersive sessions. Something like this instead?

> The `XRView`s returned must have a transform equal to the `originOffset` for a single view or the `originOffset` with an additional offset to account for stereo content with multiple views.

toji: Side effect of moving the spaces section below this one: We now reference `getPose()` here prior to defining what it does later in the doc. :roll_eyes:

You know those aliens from Arrival who's written language is non-linear and transcends the concept of time? They must write some awesome tech documents.

toji: Sorry, this isn't actually a new bug in this PR, I'm just calling it out because your change highlighted it for me. We can fix it separately.

NellWaliczek: ah good point.  I can fix it in this PR, that's fine.

NellWaliczek: Yeah, this was my biggest hesitation with moving things around... Is this better?

toji:
Looks good!

NellWaliczek:
Moved to WebXR repo.  Closing the PR, but leaving the branch for now

PR #26, opened 2019-01-18
Gamepad-based button/axis input proposal

toji:


input-explainer.md
```diff
@@ -280,6 +280,78 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
 
 For each of these events the `XRInputSource`'s target ray must be updated to originate at the point that was interacted with on the canvas, projected onto the near clipping plane (defined by the `depthNear` attribute of the `XRSession`) and extending out into the scene along that projected vector.
 
+## Button and Axis State
+
+Some applications need more than point-and-click style interaction provided by the `select` events. For devices that make use of controllers with buttons and axes, more complete information about the state of those inputs can be observed via the `XRInputSource`'s `gamepad` attribute.
+
+If the `XRInputState` represents an XR controller device with buttons or axes the `gamepad` attribute is an instance of a [`Gamepad`](https://w3c.github.io/gamepad/#gamepad-interface) interface, and `null` otherwise. Examples of controllers that may expose their state this way include Oculus Touch, Vive wands, Oculus Go and Daydream controllers, or other similar devices. More traditional gamepads, such as XBox or Playstation controllers, should not be exposed using this interface, nor should tracked devices without discreet inputs, such as optical hand tracking.
```
NellWaliczek: >More traditional gamepads, such as XBox or Playstation controllers, should not be exposed using this interface

Why not?

input-explainer.md
```diff
@@ -280,6 +280,78 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+
+`Gamepad` instances reported in this way have several notable behavioral changes vs. the ones reported by `navigator.getGamepads()`:
+
+ - `Gamepad` instances connected to an `XRInputSource` must not be included in the array returned by `navigator.getGamepads()`.
+ - The `Gamepad`'s `id` attribute must be empty string. The `XRInputState`'s `renderId` attribute should be used instead to identify the device.
```
NellWaliczek: Should we be using the `id` as the `renderId`?

input-explainer.md
```diff
@@ -280,6 +280,78 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+
+All other attributes behave as described in the [Gamepad](https://w3c.github.io/gamepad/) specification.
+
+The WebXR Device API introduces a new standard controller layout indicated by the `mapping` value of `xr-standard`. (Additional mapping variants may be added in the future if necessary.) This defines a specific layout for the inputs most commonly found on XR controller devices today. The following table describes the buttons/axes and their physical locations:
+
+| Button/Axis  | Location                |
+| ------------ | ----------------------- |
+| buttons[0]   | Primary trigger         |
+| buttons[1]   | Touchpad/Joystick click |
+| buttons[2]   | Grip/Secondary trigger  |
```
NellWaliczek: Let's list the entry for buttons[3], axes[2], and axes[3] even though we'll call them out as optional below.

input-explainer.md
```diff
@@ -280,6 +280,78 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+| axes[0]      | Touchpad/Joystick X     |
+| axes[1]      | Touchpad/Joystick Y     |
+
+Additional device-specific inputs may be exposed after these reserved indices, but devices that lack one of the canonical inputs must still preserve their place in the array. If a device has both a touchpad and a joystick the UA should designate one of them to be the primary axis-based input and expose the other at axes[2] and axes[3] with an associated button at button[4].
```
NellWaliczek: I think you meant button[3]?  Also, you never call out that axes[0] and axes[1] are associated with button[1].

input-explainer.md
```diff
@@ -280,6 +280,78 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+
+If the `XRInputState` represents an XR controller device with buttons or axes the `gamepad` attribute is an instance of a [`Gamepad`](https://w3c.github.io/gamepad/#gamepad-interface) interface, and `null` otherwise. Examples of controllers that may expose their state this way include Oculus Touch, Vive wands, Oculus Go and Daydream controllers, or other similar devices. More traditional gamepads, such as XBox or Playstation controllers, should not be exposed using this interface, nor should tracked devices without discreet inputs, such as optical hand tracking.
+
+`Gamepad` instances reported in this way have several notable behavioral changes vs. the ones reported by `navigator.getGamepads()`:
+
+ - `Gamepad` instances connected to an `XRInputSource` must not be included in the array returned by `navigator.getGamepads()`.
+ - The `Gamepad`'s `id` attribute must be empty string. The `XRInputState`'s `renderId` attribute should be used instead to identify the device.
+ - The `Gamepad`'s `index` attribute must be `-1`.
```
NellWaliczek: We could choose to say that this is only true for Gamepads with the `xr-standard` mapping if we wanted to allow the associated gamepad for `gaze` input sources...

input-explainer.md
```diff
@@ -280,6 +280,78 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+ - The `Gamepad`'s `connected` attribute must always be `true`.
+ - The `Gamepad`'s `mapping` attribute must be `xr-standard`.
+
+All other attributes behave as described in the [Gamepad](https://w3c.github.io/gamepad/) specification.
```
NellWaliczek: Unrelatedly... when do we officially kill the gamepad extensions?

input-explainer.md
```diff
@@ -280,6 +280,78 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+| buttons[1]   | Touchpad/Joystick click |
+| buttons[2]   | Grip/Secondary trigger  |
+| axes[0]      | Touchpad/Joystick X     |
+| axes[1]      | Touchpad/Joystick Y     |
+
+Additional device-specific inputs may be exposed after these reserved indices, but devices that lack one of the canonical inputs must still preserve their place in the array. If a device has both a touchpad and a joystick the UA should designate one of them to be the primary axis-based input and expose the other at axes[2] and axes[3] with an associated button at button[4].
+
+```js
+function onXRFrame(timestamp, frame) {
+  let inputSource = xrSession.getInputSources()[0];
```
NellWaliczek: This seems a bit dangerous... can we use the `primaryInputSource` variable?

input-explainer.md
```diff
@@ -280,6 +280,78 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+
+  // Check to see if the input source has buttons/axes.
+  if (inputSource.gamepad) {
+    
+    // Use joystick or touchpad values for movement.
+    MoveUser(gamepad.axes[0], gamepad.axes[0]);
```
NellWaliczek: was one of these supposed to have an index of 1?  Were you meaning to check if they were not undefined?

input-explainer.md
```diff
@@ -280,6 +280,78 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+    }
+    
+    // etc.
+  }
+
+  // Do the rest of typical frame processing...
+}
+```
+
+If the application includes interactions that require user activation (such as starting media playback), the application can listen to the `XRInputSource`s `select` events, which fire for every pressed and released button on the controller. When triggered by a controller input, the `XRInputSourceEvent` will include a `buttonIndex` other than `-1` to indicate which button on the gamepad triggered the event.
```
NellWaliczek: Honestly, I suspect this is going to be the most contentious part :-/ I'm kinda curious what it would look like to propose events on the Gamepad object itself?

input-explainer.md
```diff
@@ -280,6 +280,78 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+
+The UA may update the `gamepad` state at any point, but it must remain constant while running a batch of `XRSession` `requestAnimationFrame` callbacks.
```
NellWaliczek: And during the event callbacks as well....

NellWaliczek: Ah, which is maybe a reason not to have the traditional gamepads show up?  Though, again, we might be able to split out the behavior based on the mapping?

input-explainer.md
```diff
@@ -337,7 +409,27 @@ interface XRInputSource {
   readonly attribute XRTargetRayMode targetRayMode;
   readonly attribute XRSpace targetRaySpace;
   readonly attribute XRSpace? gripSpace;
+  readonly attribute Gamepad? gamepad;
+};
+
+//--------------------------------------------------------------------
+// FIXME: Remove before sending to Nell. For temporary reference only.
```
NellWaliczek: *snickers*

input-explainer.md
toji: Then it would only apply to controllers with a gamepad component. I'm not completely convinced the venn diagram of "Things you want to render" and "things that have a gamepad" is a circle. (Could be? I just haven't thought it through completely.)

input-explainer.md
toji: I still feel like there's generally not a clear way that most platforms associate an untracked gamepad with the headset. If the device has 4 XBox controllers plugged in I think most developers would find it weird to see all four controllers showing up as being associated with the headset somehow.

That being said, I could see updating this to instead state that "controllers not directly associated with the XR device (such as the majority of traditional gamepads) should not be exposed using this interface."

input-explainer.md
toji: As soon as WebVR is dead and buried. :) We could push for the extensions page to be deprecated right away, but it was never particularly official anyway and the most important part of it (the `touched` attribute) has already been migrated over to the spec proper.

input-explainer.md
toji: Sure

input-explainer.md
toji: Yes, the second one should be `[1]`. And if the gamepad has the `xr-standard` mapping this should always be there. (But that means I should check for the mapping.)

input-explainer.md
toji: This ancient PR still seems to be getting a surprising amount of semi-recent updates, and I know that some code related to that has been making it's way into Chrome. https://github.com/w3c/gamepad/pull/15

input-explainer.md
toji: 😱 

(You had ONE JOB, comment...)

input-explainer.md
toji: Ewe... new thought, after reading through some of the `renderId`  feedback: With a `gamepad`-based input system the assumption is that we'd end up using something like `renderId` for a lot of the more detailed mapping. But on an AR device with a gamepad controller (ie: Magic Leap) the `renderId` is likely to be empty string because you aren't supposed to render anything! So maybe this needs to duplicate the `renderId` OR be what the `renderId` would have been in cases when you shouldn't be rendering the device. Which.... ewe. :(

input-explainer.md
NellWaliczek: Ah, so in our Edge implementation, we just implicitly associated gamepads with the active XR session.  I suppose that's a bit more awkward now that we have inline sessions... we should check with Microsoft about what they want to do about that....

input-explainer.md
NellWaliczek: ugh.  uuuugh. i need to think on this a bit more...

input-explainer.md
NellWaliczek: Are we sure that there's no tracked controller with just buttons and no x/y thingie?

input-explainer.md
NellWaliczek: oof.  which is something we should think about then....

input-explainer.md
NellWaliczek: Ok, what if we always fill in renderId when there's a hardware form factor, and rely on using the environmentblendmode to determine actual rendering?

toji: Sounds pretty reasonable? Are there any cases where you'd want an input to be rendered with a non-opaque display that *aren't* app-driven decisions? (Because if they want to render a magic wand or something that doesn't occlude with your hand properly or something, well... have fun with that.)

The only scenario I can think of (and it's a stretch) is if you had a virtual or debug input that was being driven by something other than a tracked source. But... eh.... meh? :roll_eyes:

toji:
Closing post-F2F, all updated have been moved over to webxr/499

PR #27, opened 2019-01-24
Working with gamepad data

NellWaliczek:
So per our conversation the other day, I'm trying to round up a comprehensive understanding of how gamepad hardware works, the inventory of what developers need to know, and how they're likely to use the data.

This is sooooo not a proposal.  I just wanted you to see the beginnings of my capturing of the data.  I followed a very glTF style in how I've written it all down so far.  "buttonIndex" and "axisIndex" were for the mapping back to the gamepads array, but everything else should be non-duplicate data.  We might want to hop on a call this afternoon to walk through what I've got so far and figure out if you're needing something else instead.


NellWaliczek:
I filled in a bunch more text/samples so hopefully document should be a bit more self-explanatory.  And I'm generally feeling pretty good about this after having shown it to some people at my office, though I'd really love to get your take on it.
A few notes:
* As always, the names are total gibberish.
* I'm still feeling a bit iffy about the way I'm representing the touchpad's motion by splitting into a dpad motion, but that seems like small tweak if the overall direction seems right.
* I also need to go back and double check the range values for all the controllers, but at this stage of the design that didn't seem necessary.

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+        "primaryAxes" : #
+    }
+}
+```
+
+### Components
+```json
+"components" : [
+    {
+        "name" : "<friendly name>",
```
toji: If everything is expected to have a friendly name, should that act as a dictionary key instead?

Also, this bumps up against Kip's point on the other controller proposal that we need to be taking into account localization, which I don't see addressed in this doc. I would think we'd want anything that may need to be displayed to a user to be in a dictionary like:

```
"displayName": {
  "en-us": "Home Button",
  "fr-fake": "Le Homme Bouton" 
}
```

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+        "centerButton" : {
+        }
+    }
+}
+```
+### Assets
+FILL ME IN
+Note that for this section, the two rootNodes could also be the same
+```json
+"assets" : {
```
toji: Since we're breaking this down by handedness as well, shouldn't this go under the `hands` dictionary as well?

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+    "neutralHand" : {
+        "asset" : "some uri",
+        "rootNode" : "node in the asset to go at gripSpace origin"
+    },
+    "visualizationNodes" : [
+        {
+            "component" : #,
+            "rootNode" : "<node in the asset of the component's root>",
+            "labelNode" : "<node in the asset to place a usage label>",
+            "buttonMotion" : {}
```
toji: Not sure I grok what this is supposed to do?

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+"assets" : {
+    "neutralHand" : {
+        "asset" : "some uri",
+        "rootNode" : "node in the asset to go at gripSpace origin"
+    },
+    "visualizationNodes" : [
+        {
+            "component" : #,
+            "rootNode" : "<node in the asset of the component's root>",
+            "labelNode" : "<node in the asset to place a usage label>",
```
toji: Similar concern to the localization point from above.

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+```json
+"assets" : {
+    "neutralHand" : {
+        "asset" : "some uri",
+        "rootNode" : "node in the asset to go at gripSpace origin"
+    },
+    "visualizationNodes" : [
+        {
+            "component" : #,
+            "rootNode" : "<node in the asset of the component's root>",
```
toji: This may need to be per-hand, right? Unless we expect the pattern to be that both hands always use the exact same layout, which would likely require them to always be separate files, which is probably not desirable.

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+This section covers the mappings for all known XR controllers
+
+### Windows Mixed Reality
+```json
+"045E-065D" : {
+    "name" : "Windows Motion Controller",
+    "mapping" : "xr-standard",
+    "hands" : {
+        "left" : {
+            "components" : [ 0, 1, 2, 3, 4],
```
toji: I'm... not really clear on what this level of indirection buys us?

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+        "defaultValue" : 0,
+        "min" : 0,
+        "max" : 1,
+    },
+    "downButton" : {
+        "defaultValue" : 0,
+        "min" : 0,
+        "max" : 1,
+    },
+    "centerButton" : {
```
toji: Don't think I've ever seen a dpad that supports a "center" button?

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+            "defaultValue" : 0,
+            "leftValue" : -1,
+            "rightValue" : 1,
+        },
+        "yAxis" : {
+            "defaultValue" : 0,
+            "upValue" : -1,
+            "downValue" : 1,
+        },
+        "centerButton" : {
```
toji: Not sure why this would be "centerButton" instead of just "button".

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+        "min" : 0,
+        "max" : 1,
+    }
+}
+```
+
+> **OPEN ISSUE** Should we have a "switch" style button?  Or is that a subtype of something else? buttons? dpads?
+
+#### Thumbsticks
+Thumbsticks always report values in a range such that `x` is between `xAxis.leftValue` and `xAxis.rightValue` and such that `y` is between `yAxis.downValue` and `yAxis.upValue`.  There is no guarantee that `xAxis.leftValue` is less than `xAxis.rightValue`, but the values will not be equal.  There is no guarantee that `yAxis.downValue` is less than `yAxis.upValue`, but the values will not be equal.
```
toji: The range for the various axes should be something that is determined by the spec. The gamepad API, for example, requires that all axis values are normalized into a `[-1, 1]` range. (And if we don't use gamepad directly we should encode a similar requirement into our own spec.) As such, this would effectively always report the same values unless the UA's implementation is outright broken.

Same goes for the button range above.

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+        "supportsPress" : true,
+        "analogValues" : false,
+        "defaultValue" : 0,
+        "min" : 0,
+        "max" : 1,
+    }
+}
+```
+
+> **OPEN ISSUE** Should we have a "switch" style button?  Or is that a subtype of something else? buttons? dpads?
```
toji: A switch would just be a button that persists in the "pressed" state longer than usual, right?

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+
+## Gamepads
+FILL ME IN
+```json
+{
+    "gamepads" : [
+        {
+            "<id of gamepad>" : {
+                "name" : "<friendly name of gamepad>",
+                "mapping" : "xr-standard",
```
toji: What's the value of stating this here rather than reading it off the gamepad?

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+        }
+    ]
+}
+```
+
+#### Motion types
+FILL ME IN
+
+## User-agent overrides
+FILL ME IN - and design me
```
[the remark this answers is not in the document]
toji: Oh, and if the above is correct, we need to make sure that it accounts for version ranges of a given UA. 😝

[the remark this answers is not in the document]
NellWaliczek: Yep, that's the idea.  And also, yes :)

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+  * Is there a button associated with the axes?
+* priority ordering of buttons (which touchpad/thumbstick matters)
+* Where is each button relative to the gripSpace?
+  * for putting a “how do you use me” label
+* Does the button move? 
+  * Is there bone ID?
+  * What are the directions/limits of motion?
+  * Touchpad location indicator?
+* Still to poke at 
+  * Oculus Go and Touch fires events at the extents of the motion controller bounds.
```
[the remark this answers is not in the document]
NellWaliczek: No sorry, crappy note to myself.  It was to figure out how to handle the fact that these devices, and the vive too, have "button" behavior when you press far left, right, up, and down on the trackpads.  This is why it's dpad-like.

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+  * Vive seems to also treat trackpad regions as "being buttons"
+  * figure out how to represent trackpad edge clicks
+
+# Schema explanation
+FILL ME IN
```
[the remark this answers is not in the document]
NellWaliczek: -_- you think? :p

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+```
+
+### Hands
+For motion controllers that report a handedness property
+```json
+"hands" : {
+    "left" : {
+        "components" : [#],
+        "primaryButton" : #,
+        "primaryAxes" : #
```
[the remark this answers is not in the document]
NellWaliczek: I put the "#" to indicate that it was an index into the components array.  I'm following a glTF-style here.

gamepad-input-explainer.md
NellWaliczek: Yes, however, I wanted to leave a tool to handle bugs. So perhaps this should be only in the UA overrides section....

gamepad-input-explainer.md
NellWaliczek: That's kinda what I was thinking.  Which is why I didn't create a new physical type for it.  Ok, I'll remove the comment since you're thinking the same thing...

gamepad-input-explainer.md
NellWaliczek: Ah, this is a leftover from when I was considering including the HoloLens clicker and Oculus Remote in this mapping.  That said, the data will be on the gamepad object, so you're right it can absolutely be removed.

gamepad-input-explainer.md
```diff
@@ -0,0 +1,1077 @@
...
+FILL ME IN
+```json
+{
+    "gamepads" : [
+        {
+            "<id of gamepad>" : {
+                "name" : "<friendly name of gamepad>",
+                "mapping" : "xr-standard",
+                "hands" : {},
+                "components" : [],
```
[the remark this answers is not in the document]
NellWaliczek: > If the hands dictionary can account for non-handed gamepads, then does the components array here serve a purpose?

I'm not entirely sure I understand this question.  What are you suggesting instead?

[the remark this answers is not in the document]
NellWaliczek: Also, are you bikeshedding?? :p

gamepad-input-explainer.md
NellWaliczek: And by "just yet" I meant in the first draft.  We could add a section for it before we share this with the group, though that seems like a lot all at once... I think separate PR for it would pretty reasonable.

gamepad-input-explainer.md
NellWaliczek: I very purposely separated the sections by "data that comes from the gamepad api" and "things we'll use that data for".  The approach was inspired by glTF, tbh.

gamepad-input-explainer.md
NellWaliczek: Not sure what this has to do with loc?  This is just an empty node in the asset that indicates to an engine where a safe place is to put a usage label if needed.  For example if a touchpad is used for scrolling, the engine might want to show a "scroll" label.  This is necessary because the goal of this design is to make the usage universal (and not depend on actually knowing exactly which motion controller is being used)

gamepad-input-explainer.md
NellWaliczek: This allows the engine to animate the asset based on the values that come off the associated component.  Without needing to know which exact brand of motion controller it is.  This is the whole point of the design :)

gamepad-input-explainer.md
NellWaliczek: I don't understand this question... that's the purpose of the "hands" node in the "assets" section...

gamepad-input-explainer.md
NellWaliczek: Haha this is the same answer as your previous question.  We should talk through this when we chat later today.

NellWaliczek:
This has been moved to it's own repo: immersive-web/gamepad-mappings

PR #29, opened 2019-02-05
Spec text for the identity reference space

toji:
First pass at adding the `"identity"` reference space type. Not including `originOffset` at this point to keep PRs focused and easy to review. PTAL!


toji:
Moved to webxr/520

PR #31, opened 2019-02-28
Add the ignoreDepthValues attribute to the XRWebGLLayer

toji:
(Something goofed up in the previous pull request and I couldn't get it to ignore a commit from Blair that was already in master.)

Addresses webxr/523. Given the scope of the feature it felt appropriate to handle both the spec text and explainer text in the same PR. Please take a look!


explainer.md
```diff
@@ -520,6 +520,16 @@ xrSession.updateRenderState({
 });
 ```
 
+### Preventing the compositor from using the depth buffer
+
+By default the depth attachment of an `XRWebGLLayer`'s `framebuffer`, if present, may be used to assist the XR compositor. For example, some advanced reprojection techniques make use of the scene's depth values. This assumes, of course, that the values in the depth buffer are representative of the scene content.
```
NellWaliczek: I'd love to add a sentence about how this data might also be used in the future (perhaps even to get DOM layers working right, though I wouldn't want to mention that directly).

explainer.md
```diff
@@ -520,6 +520,16 @@ xrSession.updateRenderState({
...
+
+Some applications may violate that assumption, such as when using certain deferred rendering techniques or rendering stereo video. In those cases if the depth buffer's values are used by the compositor it may result in objectionable artifacts. To avoid this, the compositor can be instructed to ignore the depth values of an `XRWebGLLayer` by setting the `ignoreDepthValues` option to `true` at layer creation time:
```
NellWaliczek: We might want to call out that there would be a downside to taking this approach in terms of not being able to use certain features.

explainer.md
```diff
@@ -689,6 +699,7 @@ dictionary XRWebGLLayerInit {
   boolean depth = true;
   boolean stencil = false;
   boolean alpha = true;
+  boolean ignoreDepthValues = false;
```
NellWaliczek: Seeing this in context, I'm wondering if it gets slightly confusing with the "depth" property.

index.bs
```diff
@@ -1175,6 +1178,8 @@ The <dfn attribute for="XRWebGLLayer">stencil</dfn> attribute is <code>true</cod
 
 The <dfn attribute for="XRWebGLLayer">alpha</dfn> attribute is <code>true</code> if the {{framebuffer}} has an alpha buffer attachment and <code>false</code> if no alpha buffer is attached.
 
+The <dfn attribute for="XRWebGLLayer">ignoreDepthValues</dfn> attribute, if <code>true</code>, indicated the [=XR Compositor=] MUST NOT make use of values in the depth buffer attachment when rendering. When the attribute is <code>false</code> it indicates that the content of the depth buffer attachment is expected to be representative of the scene rendered into the layer and the compositor MAY use it for effects such as improved reprojection.
```
NellWaliczek: nit: indicated -> indicates

explainer.md
toji: Did  you have something in mind in terms of future uses? I've added some language around how the UA is free to use it for whatever it sees fit, and also called out an example of compositing platform and UA interfaces without depth disparity problems, but it kinda sounds like you had something else in mind?

explainer.md
toji: The same thought occurred to me, but I have a hard time seeing how ANY descriptor without the word "depth" in it (which is the real point of confusion here) is going to communicate the appropriate thing.

toji:
Addressed feedback and moved to webxr/548

PR #33, opened 2019-04-19
[Editorial] Status metadata does not need "w3c/"

himorin:
Status metadata for bikeshed does not require "w3c/". This will also fix missing banner at top left for editors' draft.


toji:
We typically do not merge PRs in editor-collab, but instead merge fixes in the /webxr repo and then push the updated master to this repo (which is this case has already been done with your PR in that repo.) Thank you for your contribution, though!

PR #36, opened 2019-04-23
Formalize origin definitions and some transform maths

toji:
A better fleshed out version of some of #35 to facilitate discussion with Klaus and Manish.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -670,7 +670,7 @@ An {{XRFrame}} represents a snapshot of the state of all of the tracked objects 
   [SameObject] readonly attribute XRSession session;
 
   XRViewerPose? getViewerPose(XRReferenceSpace referenceSpace);
-  XRPose? getPose(XRSpace space, XRSpace relativeTo);
+  XRPose? getPose(XRSpace sourceSpace, XRSpace destinationSpace);
 };
 </pre>
 
@@ -678,19 +678,29 @@ Each {{XRFrame}} has a <dfn for="XRFrame">active</dfn> boolean which is initiall
 
 The <dfn attribute for="XRFrame">session</dfn> attribute returns the {{XRSession}} that produced the {{XRFrame}}.
 
+<div class="algorithm" data-algorithm="valid-pose">
+
+To determine if {{XRFrame}} |frame| can evaluate a <dfn>valid pose</dfn> between {{XRSpace}}s |sourceSpace| and |destinationSpace|, the user agent MUST run the following steps:
+
+1. If |frame|'s [=active=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
+1. Let |session| be |frame|'s {{XRFrame/session}} object.
+1. If |sourceSpace|'s [=XRSpace/session=] does not equal |session| return <code>false</code>.
+1. If |destinationSpace|'s [=XRSpace/session=] does not equal |session| return <code>false</code>.
+1. Return <code>true</code>.
+
+</div>
+
 <div class="algorithm" data-algorithm="get-viewer-pose">
 
 When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
 
   1. Let |frame| be the target {{XRFrame}}
-  1. If |frame|'s [=active=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
-  1. If |frame|'s [=animationFrame=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
   1. Let |session| be |frame|'s {{XRFrame/session}} object.
-  1. If |referenceSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
-  1. Let |transform| be the [=transform between=] |session|'s {{XRSession/viewerSpace}} and |referenceSpace| at the time represented by |frame|.
-  1. If |transform| is <code>null</code> return <code>null</code>.
+  1. If |frame|'s [=animationFrame=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
+  1. If |frame| cannot evaluate a [=valid pose=] between |session|'s {{XRSession/viewerSpace}} and |referenceSpace|, throw an {{InvalidStateError}} and abort these steps.
   1. Let |pose| be a new {{XRViewerPose}} object.
-  1. Set |pose|'s {{XRPose/transform}} to |transform|.
+  1. Set |pose|'s {{XRPose/transform}} to the [=transform from=] |session|'s {{XRSession/viewerSpace}} to |referenceSpace| at the time represented by |frame|.
+  1. If |pose|'s {{XRPose/transform}} is <code>null</code> return <code>null</code>.
   1. Return |pose|.
 
 </div>
@@ -702,14 +712,10 @@ ISSUE: Describe how the {{XRViewerPose/views}} are populated.
 When the <dfn method for="XRFrame">getPose(|sourceSpace|, |destinationSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
 
   1. Let |frame| be the target {{XRFrame}}
-  1. If |frame|'s [=active=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
-  1. Let |session| be |frame|'s {{XRFrame/session}} object.
-  1. If |sourceSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
-  1. If |destinationSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
-  1. Let |transform| be the [=transform between=] |sourceSpace| and |destinationSpace| at the time represented by |frame|.
-  1. If |transform| is <code>null</code> return <code>null</code>.
+  1. If |frame| cannot evaluate a [=valid pose=] between |sourceSpace| and |destinationSpace|, throw an {{InvalidStateError}} and abort these steps.
   1. Let |pose| be a new {{XRPose}} object.
-  1. Set |pose|'s {{XRPose/transform}} to |transform|.
+  1. Set |pose|'s {{XRPose/transform}} to the [=transform from=] |sourceSpace| to |destinationSpace| at the time represented by |frame|.
+  1. If |pose|'s {{XRPose/transform}} is <code>null</code> return <code>null</code>.
   1. Return |pose|.
 
 </div>
@@ -732,18 +738,18 @@ An {{XRSpace}} describes an entity that is tracked by the [=/XR device=]'s track
 
 Each {{XRSpace}} has a <dfn for="XRSpace">session</dfn> which is set to the {{XRSession}} that created the {{XRSpace}}.
 
-Each {{XRSpace}} has a <dfn for="XRSpace">native origin</dfn> that describes the origin of the {{XRSpace}} relative to a component of the [=/XR device=]'s tracking systems or features of the user's physical environment. Each {{XRSpace}} also has an <dfn for="XRSpace">origin offset</dfn> which is an {{XRRigidTransform}}, initially set to an [=identity transform=] that describes an offset from the [=native origin=].
+Each {{XRSpace}} has a <dfn for="XRSpace">native origin</dfn> which is an {{XRRigidTransform}} that describes the origin of the {{XRSpace}} relative to a component of the [=/XR device=]'s tracking systems or features of the user's physical environment. Each {{XRSpace}} also has an <dfn for="XRSpace">origin offset</dfn> which is an {{XRRigidTransform}}, initially set to an [=identity transform=] that describes an offset from the [=native origin=].
 
 To determine the <dfn for="XRSpace">effective origin</dfn> of an {{XRSpace}}, [=multiply=] the [=native origin=] by the [=origin offset=]. The [=effective origin=] is the basis of the <dfn for="XRSpace">coordinate system</dfn> described by the {{XRSpace}}.
 
-<div class="algorithm" data-algorithm="transform-between-spaces">
+<div class="algorithm" data-algorithm="transform-from-to">
 
-To <dfn lt="transform between">get the transform between two spaces</dfn>, a |sourceSpace| and a |destinationSpace|, the user agent MUST run the following steps:
+To <dfn lt="transform from">get the transform from</dfn> a |sourceSpace| to a |destinationSpace|, the user agent MUST run the following steps:
 
   1. If |destinationSpace|'s pose cannot be determined relative to |sourceSpace|, return <code>null</code>
   1. Let |transform| be a new {{XRRigidTransform}} object.
-  1. Set |transform|'s {{XRRigidTransform/position}} to the location of |destinationSpace|'s [=effective origin=] in |sourceSpace|'s [=coordinate system=].
-  1. Set |transform|'s {{XRRigidTransform/orientation}} to the rotation of |destinationSpace|'s [=effective origin=] in |sourceSpace|'s [=coordinate system=].
+  1. Set |transform|'s {{XRRigidTransform/position}} to the location of |sourceSpace|'s [=effective origin=] in |destinationSpace|'s [=coordinate system=].
+  1. Set |transform|'s {{XRRigidTransform/orientation}} to the rotation of |sourceSpace|'s [=effective origin=] in |destinationSpace|'s [=coordinate system=].
   1. Return |transform|.
 
 </div>
--- a/index.bs
+++ b/index.bs
@@ -728,7 +728,9 @@ Spaces {#spaces}
 XRSpace {#xrspace-interface}
 ------------------
 
-An {{XRSpace}} describes an entity that is tracked by the [=/XR device=]'s tracking systems. {{XRSpace}}s MAY NOT have a fixed spatial relationship to one another or to any given {{XRReferenceSpace}}. The transform between two {{XRSpace}} can be evaluated by calling an {{XRFrame}}'s {{XRFrame/getPose()}} method. The interface is intentionally opaque.
+An {{XRSpace}} describes an entity that is tracked by the [=/XR device=]'s tracking systems. Any time a position, orientation, or direction is input or output with the WebXR Device API, the {{XRSpace}} those values are defined in must be specified.
+
+{{XRSpace}}s MAY NOT have a fixed spatial relationship to one another or to any given {{XRReferenceSpace}}. The transform between two {{XRSpace}} can be evaluated by calling an {{XRFrame}}'s {{XRFrame/getPose()}} method. The interface is intentionally opaque.
 
 <pre class="idl">
 [SecureContext, Exposed=Window] interface XRSpace : EventTarget {
@@ -738,9 +740,11 @@ An {{XRSpace}} describes an entity that is tracked by the [=/XR device=]'s track
 
 Each {{XRSpace}} has a <dfn for="XRSpace">session</dfn> which is set to the {{XRSession}} that created the {{XRSpace}}.
 
-Each {{XRSpace}} has a <dfn for="XRSpace">native origin</dfn> which is an {{XRRigidTransform}} that describes the origin of the {{XRSpace}} relative to a component of the [=/XR device=]'s tracking systems or features of the user's physical environment. Each {{XRSpace}} also has an <dfn for="XRSpace">origin offset</dfn> which is an {{XRRigidTransform}}, initially set to an [=identity transform=] that describes an offset from the [=native origin=].
+Each {{XRSpace}} has a <dfn for="XRSpace">native origin</dfn> which is an {{XRRigidTransform}} from the native origin's space to the [=/XR device=]'s underlying tracking system and the user's physical environment. 
 
-To determine the <dfn for="XRSpace">effective origin</dfn> of an {{XRSpace}}, [=multiply=] the [=native origin=] by the [=origin offset=]. The [=effective origin=] is the basis of the <dfn for="XRSpace">coordinate system</dfn> described by the {{XRSpace}}.
+Each {{XRSpace}} also has an <dfn for="XRSpace">origin offset</dfn> which is an {{XRRigidTransform}}, from the effective space to the [=native origin=]'s space. It is initially set to an [=identity transform=].
+
+To determine the <dfn for="XRSpace">effective origin</dfn> of an {{XRSpace}}, [=multiply transforms|multiply=] the [=native origin=] by the [=origin offset=]. The [=effective origin=] is the basis of the <dfn for="XRSpace">coordinate system</dfn> described by the {{XRSpace}}.
 
 <div class="algorithm" data-algorithm="transform-from-to">
 
@@ -960,9 +964,20 @@ Geometric Primitives {#geometricprimitives}
 Matrices {#matrices}
 --------
 
-WebXR provides various transforms in the form of <dfn lt="matrix">matrices</dfn>. WebXR matrices are always 4x4 and given as 16 element {{Float32Array}}s in column major order. They may be passed directly to WebGL's {{uniformMatrix4fv}} function, used to create an equivalent {{DOMMatrix}}, or used with a variety of third party math libraries.
+WebXR provides various transforms in the form of <dfn lt="matrix">matrices</dfn>. WebXR uses the WebGL conventions when communicating matrices, in which 4x4 matrices are given as 16 element {{Float32Array}}s with column major storage and column vectors. They may be passed directly to WebGL's {{uniformMatrix4fv}} function, used to create an equivalent {{DOMMatrix}}, or used with a variety of third party math libraries.
 
-Translations specified by WebXR matrices are always given in meters.
+<div class="example">
+For a 16 element {{Float32Array}} such as this:
+<pre>[a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p]</pre>
+
+The corresponding matrix is understood to be laid out like so:
+<pre>
+a e i m
+b f j n
+c g k o
+d h l p
+</pre>
+</div>
 
 Normalization {#normalization}
 -------------
@@ -1022,13 +1037,15 @@ An {{XRRigidTransform}} with a {{XRRigidTransform/position}} of <code>{ x: 0, y:
 
 <div class="algorithm" data-algorithm="multiply transforms">
 
-To <dfn lt="multiply">multiply two {{XRRigidTransform}}s</dfn>, |first| and |second|, the UA MUST perform the following steps:
+To <dfn lt="multiply transforms">multiply two {{XRRigidTransform}}s</dfn>, |B| and |A|, the UA MUST perform the following steps:
 
   1. Let |result| be a new {{XRRigidTransform}} object.
-  1. Set |result|'s {{XRRigidTransform/matrix}} to the result of multiplying |first|'s {{XRRigidTransform/matrix}} by |second|'s {{XRRigidTransform/matrix}}.
+  1. Set |result|'s {{XRRigidTransform/matrix}} to the result of premultiplying |B|'s {{XRRigidTransform/matrix}} from the left onto |A|'s {{XRRigidTransform/matrix}}.
   1. Set |result|'s {{XRRigidTransform/orientation}} to the quaternion that describes the rotation indicated by the top left 3x3 sub-matrix of |result|'s {{XRRigidTransform/matrix}}.
   1. Set |result|'s {{XRRigidTransform/position}} to the vector given by the fourth column of |result|'s {{XRRigidTransform/matrix}}.
   1. Return |result|.
+
+|result| is a transform from |A|'s source space to |B|'s destination space.
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -712,7 +712,7 @@ A core feature of the WebXR Device API is the ability to provide spatial trackin
 XRSpace {#xrspace-interface}
 ------------------
 
-An {{XRSpace}} represents a virtual coordinate system that corresponds to a physical location. Any time spatial data is requested from the API or given to the API, the {{XRSpace}} it is expressed in must be specified. The interface is intentionally opaque.
+An {{XRSpace}} represents a virtual coordinate system with an origin that corresponds to a physical location. Spatial data that is requested from the API or given to the API is always expressed in relation to a specific {{XRSpace}}, and numeric values such as pose positions are coordinates in that space relative to its origin. The interface is intentionally opaque.
 
 <pre class="idl">
 [SecureContext, Exposed=Window] interface XRSpace : EventTarget {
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -682,28 +682,40 @@ The <dfn attribute for="XRFrame">session</dfn> attribute returns the {{XRSession
 
 When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
 
-  1. If the {{XRFrame}}'s [=active=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
-  1. If the {{XRFrame}}'s [=animationFrame=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
-  1. Let |session| be the {{XRFrame}}'s {{XRFrame/session}} object.
+  1. Let |frame| be the target {{XRFrame}}
```
NellWaliczek: I'd love it if we could move a bunch of this common text into a common "helper" algorithm.

index.bs
```diff
@@ -682,28 +682,40 @@ The <dfn attribute for="XRFrame">session</dfn> attribute returns the {{XRSession
...
+  1. If |frame|'s [=active=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
+  1. If |frame|'s [=animationFrame=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
+  1. Let |session| be |frame|'s {{XRFrame/session}} object.
   1. If |referenceSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
-  1. If the [=viewer=]'s pose cannot be determined relative to |referenceSpace|, return <code>null</code>
-  1. Return a new {{XRViewerPose}} describing the [=viewer=]'s pose relative to the origin of |referenceSpace| at the timestamp of the {{XRFrame}}.
+  1. Let |transform| be the [=transform between=] |session|'s {{XRSession/viewerSpace}} and |referenceSpace| at the time represented by |frame|.
```
NellWaliczek: Can we throw together a quick PR with this as the base branch in which we define the `viewerSpace` even if it doesn't fully refactor the spec text so that the views as being managed by it?

index.bs
```diff
@@ -682,28 +682,40 @@ The <dfn attribute for="XRFrame">session</dfn> attribute returns the {{XRSession
...
+  1. If |transform| is <code>null</code> return <code>null</code>.
+  1. Let |pose| be a new {{XRViewerPose}} object.
+  1. Set |pose|'s {{XRPose/transform}} to |transform|.
+  1. Return |pose|.
 
 </div>
 
+ISSUE: Describe how the {{XRViewerPose/views}} are populated.
```
NellWaliczek: I know you're trying to scope this as a first PR (which I TOTALLY support), but I think it would be very instructive to at least have a separate draft PR in which this is covered. It really helps to explain how this approach can be used to build out the spec text more concretely.

index.bs
```diff
@@ -682,28 +682,40 @@ The <dfn attribute for="XRFrame">session</dfn> attribute returns the {{XRSession
...
+  1. If |destinationSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
+  1. Let |transform| be the [=transform between=] |sourceSpace| and |destinationSpace| at the time represented by |frame|.
+  1. If |transform| is <code>null</code> return <code>null</code>.
+  1. Let |pose| be a new {{XRPose}} object.
+  1. Set |pose|'s {{XRPose/transform}} to |transform|.
+  1. Return |pose|.
 
 </div>
 
+ISSUE: For both pose algorithms describe how the {{XRPose/emulatedPosition}} boolean is determined.
```
NellWaliczek: I'm ok with holding on this while we land Alex's explainer text changes :)

index.bs
```diff
@@ -720,6 +732,22 @@ An {{XRSpace}} describes an entity that is tracked by the [=/XR device=]'s track
 
 Each {{XRSpace}} has a <dfn for="XRSpace">session</dfn> which is set to the {{XRSession}} that created the {{XRSpace}}.
 
+Each {{XRSpace}} has a <dfn for="XRSpace">native origin</dfn> that describes the origin of the {{XRSpace}} relative to a component of the [=/XR device=]'s tracking systems or features of the user's physical environment. Each {{XRSpace}} also has an <dfn for="XRSpace">origin offset</dfn> which is an {{XRRigidTransform}}, initially set to an [=identity transform=] that describes an offset from the [=native origin=].
```
NellWaliczek: Do we need to define what the native origin's type is?

Manishearth:
I can handle these:

> * Defining the viewerSpace more concretely along with the impact on getViewerPose() flow
> 
> * Removing the ability to dynamically change the `originOffset` on an `XRReferenceSpace`

I _think_ I can handle this but my PR may reflect an incorrect understanding

> * Detailing the behavior of the natural origin of the identity reference space as well as which other spaces it can be related to (i.e. only the XRInputSource.targetRaySpace of an screen-space input source?)

toji:
Addressed some feedback from Nell's comments as well as some of the discussion on today's call. Specifically: 

 - Broke some of the validation in the pose getters out into a separate helper algorithm.
 - Specified that the native origin is considered a XRRigidTransform (though that's a bit of a weird one, if you ask me. What's it a transform from? At least the math is more conceptually straightforward this way.)
 - Renamed the "transform between" algorithm to "transform from, to".
 - Flipped the logic in that algorithm that Klaus pointed out I had exactly backwards. 😅

index.bs
toji: Done, though I'm not completely convinced that it removed enough logic from the individual algorithms to be a net positive?

index.bs
toji: Thanks for this suggestion! I've already updated the PR to take it into account, and I'd appreciate a sanity check on whether or not it's communicating the right thing.

Manishearth:
Opened #37, #38, #39 to fix most of the points brought up by @NellWaliczek

Manishearth:
Also opened https://github.com/immersive-web/editor-collab/pull/40 to provide an alternate way to multiply rigid transforms

Manishearth:
> - Fixing the way we return the XRBoundedReferenceSpace bounds so that they're based on the effective origin

#41

index.bs
Manishearth: https://github.com/immersive-web/editor-collab/pull/42

index.bs
```diff
@@ -986,6 +1020,18 @@ The <dfn attribute for="XRRigidTransform">inverse</dfn> attribute returns a {{XR
 
 An {{XRRigidTransform}} with a {{XRRigidTransform/position}} of <code>{ x: 0, y: 0, z: 0 w: 1 }</code> and an {{XRRigidTransform/orientation}} of <code>{ x: 0, y: 0, z: 0, w: 1 }</code> is known as an <dfn>identity transform</dfn>.
 
+<div class="algorithm" data-algorithm="multiply transforms">
+
+To <dfn lt="multiply">multiply two {{XRRigidTransform}}s</dfn>, |first| and |second|, the UA MUST perform the following steps:
+
+  1. Let |result| be a new {{XRRigidTransform}} object.
+  1. Set |result|'s {{XRRigidTransform/matrix}} to the result of multiplying |first|'s {{XRRigidTransform/matrix}} by |second|'s {{XRRigidTransform/matrix}}.
```
klausw: While this is internally consistent, it's a bit surprising since the resulting transform is equivalent to applying |second| first, and then applying transform |first| on top of that as the second transform.

Can you flip |first| and |second| here and in the "multiply two XRRigidTransforms" definition above?

Optionally, if you want to be a bit clearer, specifically say that you're premultiplying |second| on the left of |first|?

Manishearth: This is a confusion that will always exist: the "natural" multiplication order when read left to right is the opposite of its transform order. Either we talk in terms of transform order, or we talk in terms of multiplication, trying to appease both simultaneously and implicitly is not productive. If this change is made, it becomes clearer transform order wise, but now "multiply A and B" becomes `B*A` which is just as surprising.

I think Nell mentioned something similar in today's meeting: ultimately this is a trade-off where neither option is universally clearer. I think the compromise of talking about multiplication but using ordering consistent with transforms will just muddy the waters further.

(This could be resolved by using row vectors but as I mentioned in the meeting all web specs use column vectors)

I definitely think this code should explicitly specify pre/post row/column (i.e say "premultiplying first on the left of second" and maybe mention at the top that this is all column vectors notation) but I don't think we should flip this.

index.bs
```diff
@@ -986,6 +1020,18 @@ The <dfn attribute for="XRRigidTransform">inverse</dfn> attribute returns a {{XR
...
+  1. Set |result|'s {{XRRigidTransform/orientation}} to the quaternion that describes the rotation indicated by the top left 3x3 sub-matrix of |result|'s {{XRRigidTransform/matrix}}.
+  1. Set |result|'s {{XRRigidTransform/position}} to the vector given by the fourth column of |result|'s {{XRRigidTransform/matrix}}.
```
klausw: Optional, and assuming we're describing |second| multiplied by |first|, aka |second * first|.matrix = |second|.matrix * |first|.matrix, we could be more specific here:

The result's position equals |first|.position rotated by |second|.orientation and added to |second|.position.

(Proof sketch, using A for |first| and B for |second| - each matrix consisting of a rotation left-multiplied by a translation, this follows from `C = B * A` => `C = B_trans * B_rot * A_trans * A_rot` applied to a `[0, 0, 0, 1]` vector.)

index.bs
```diff
@@ -986,6 +1014,18 @@ The <dfn attribute for="XRRigidTransform">inverse</dfn> attribute returns a {{XR
 
 An {{XRRigidTransform}} with a {{XRRigidTransform/position}} of <code>{ x: 0, y: 0, z: 0 w: 1 }</code> and an {{XRRigidTransform/orientation}} of <code>{ x: 0, y: 0, z: 0, w: 1 }</code> is known as an <dfn>identity transform</dfn>.
 
+<div class="algorithm" data-algorithm="multiply transforms">
+
+To <dfn lt="multiply">multiply two {{XRRigidTransform}}s</dfn>, |first| and |second|, the UA MUST perform the following steps:
```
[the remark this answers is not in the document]
klausw: > The result is a transform from |first|'s source space to |second|'s destination space.

That is only true when flipping |first| and |second| as I suggested just now in https://github.com/immersive-web/editor-collab/pull/36/files#r277968297 .

(Yes, this is confusing...)

index.bs
Manishearth: That's not quite correct, `B_rot * A_trans` is not "A's translation rotated by B` since the resultant matrix has both position and rotation components. What you're looking for is `B_rot * A_trans * B_rot^-1`, which you can obtain with some math, but the answer is definitely a bit more complex . See #40 for the algorithm (I can write down the full proof if necessary)

klausw:
Apologies for adding to the confusion with incorrect math, I shouldn't be
doing this by handwaving late at night.

Given the complications, I'm wondering if the definition of general
multiplication is really necessary if it's only used for originOffset in
the spec - maybe a less general definition may be clearer if it can refer
specifically to underlying spaces? AFAIK there's no plan to actually
introduce an XRRigidTransform multiply method at the API level, right?

Manishearth:
While writing at least two of the follow-up PRs o found the matrix multiplication

index.bs
klausw: I think we had a bit of terminology confusion here. After sleeping on it, I'm pretty sure my description is correct in terms of rotations, i.e. when considering a rotation matrix, or when treating the operation of applying a rotation quaternion to a vector as an atomic step.

Your version is correct in terms of quaternion multiplications, but the sequence `v' = q * v * q^-1` expresses the single operation of rotating vector `v` by quaternion `q`. More here, including a multiplied-out matrix to support my interpretation: https://github.com/immersive-web/editor-collab/pull/40/files#r278225851, or see the [Wikipedia entry](https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation#Using_quaternion_as_rotations).

index.bs
klausw: > This is a confusion that will always exist: the "natural" multiplication order when read left to right is the opposite of its transform order. Either we talk in terms of transform order, or we talk in terms of multiplication, trying to appease both simultaneously and implicitly is not productive. If this change is made, it becomes clearer transform order wise, but now "multiply A and B" becomes B*A which is just as surprising.

Yes, but I'm proposing that we should say "multiply B and A" which becomes B*A. (Or "multiply |second| and |first|" which becomes `|second|.matrix * |first|.matrix`. We need to keep the order consistent in any case, and I think we have some flexibility here.

More concretely, how about the following text:

```
To <dfn lt="multiply">multiply two {{XRRigidTransform}}s</dfn>, |B| and |A|, the UA MUST perform the following steps:
1. Set |result|'s {{XRRigidTransform/matrix}} to the result of premultiplying |B|'s {{XRRigidTransform/matrix}} from the left onto |A|'s {{XRRigidTransform/matrix}}.
```

> While writing at least two of the follow-up PRs I found the matrix multiplication algorithm to be _very_ useful to refer to in the spec (as opposed to using a new ad hoc way of composing transforms each time)

If it's useful, we should of course keep it. I just wanted to make sure we're not getting sidetracked.

index.bs
Manishearth: I think the confusion here is: Is `A_trans` a translation column vector, or a 4x4 translation matrix? If the former then `B_rot * A_trans` is indeed "`A_trans` rotated by `B_rot`", however you can't use that directly when talking about `B * A` since that's completely in 4x4 matrix notation.

But yes, your statement that "The result's position equals |first|.position rotated by |second|.orientation and added to |second|.position." is correct, sorry.

index.bs
Manishearth: Oh, I see, you're doing away with "first" and "second". That does make sense to me!

index.bs
Manishearth: (The source of the confusion here was that I'd done my matrix math in a correct but _slightly_ inefficient way, leading to an extra unnecessary step. I fixed this, and I think we're on the same page now)

index.bs
toji: I have no attachment to the "describes" verbiage. I'll admit that the way you lay things out above ("a transform from native origin space to underlying tracking system / physical environment space)  sounds backwards to me, though, and your comments elsewhere make me question of this is the most natural way of describing these spatial relationships for you, or if it's something that you're stretching to fit the `originOffset` decision made at the previous F2F?

If the former then I'll happily work on adapting my PR to the descriptions and verbiage you outline above. If the latter, then I'd ask how you would see these concepts working out most naturally regardless of the `originOffset` decision, because that was such a close vote that I have no qualms about overturning it should it lead to clearer spec text and implementations. Either way I want to confirm that we're not twisting ourselves in knots to work around an easily changed convention.

index.bs
```diff
@@ -720,6 +732,22 @@ An {{XRSpace}} describes an entity that is tracked by the [=/XR device=]'s track
 
 Each {{XRSpace}} has a <dfn for="XRSpace">session</dfn> which is set to the {{XRSession}} that created the {{XRSpace}}.
 
+Each {{XRSpace}} has a <dfn for="XRSpace">native origin</dfn> that describes the origin of the {{XRSpace}} relative to a component of the [=/XR device=]'s tracking systems or features of the user's physical environment. Each {{XRSpace}} also has an <dfn for="XRSpace">origin offset</dfn> which is an {{XRRigidTransform}}, initially set to an [=identity transform=] that describes an offset from the [=native origin=].
+
+To determine the <dfn for="XRSpace">effective origin</dfn> of an {{XRSpace}}, [=multiply=] the [=native origin=] by the [=origin offset=]. The [=effective origin=] is the basis of the <dfn for="XRSpace">coordinate system</dfn> described by the {{XRSpace}}.
```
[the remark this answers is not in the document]
toji: > Is the native origin roughly an XRRigidTransform from native origin space to tracking system / physical environment space? It seems to be an implementation internal feature that isn't exposed through JavaScript APIs.

Yeah, sorry, this concept is messing me up. The "native origin", as Nell and I were thinking about it, would be either the origin of the underlying tracking system or some emulated origin on top of that, as would be the case for something like a floor-level origin in an eye-level underlying system. It is definitely an implementation detail, though, and not one that should really be surfed through JavaScript.

> My understanding was that XRSpace is an abstract entity that has an origin location and orientation corresponding to abstract XYZ basis vectors. The XRSpace itself doesn't have any intrinsic numeric coordinate values on its own, but you can use it to get coordinate values for other entities relative to that space's basis.

Yes, this is my intent precisely. I'm simply unsure about how to communicate the logic that goes into establishing poses in that space. Referring to the origin as a transform is a (probably misguided) attempt to reason about that concept with primitives already defined in the spec.

> I'm assuming that we want "multiply transform X and Y" to correspond to multiplying X.matrix times Y.matrix in that order - if not, this would be rather confusing to developers.

*/vigorous head nodding* 

> If you want to do a multiplication here, I think you need a setup something like this...

I have to admit that this example lost me.

index.bs
```diff
@@ -720,6 +732,22 @@ An {{XRSpace}} describes an entity that is tracked by the [=/XR device=]'s track
...
+
+<div class="algorithm" data-algorithm="transform-between-spaces">
+
+To <dfn lt="transform between">get the transform between two spaces</dfn>, a |sourceSpace| and a |destinationSpace|, the user agent MUST run the following steps:
+
+  1. If |destinationSpace|'s pose cannot be determined relative to |sourceSpace|, return <code>null</code>
+  1. Let |transform| be a new {{XRRigidTransform}} object.
+  1. Set |transform|'s {{XRRigidTransform/position}} to the location of |destinationSpace|'s [=effective origin=] in |sourceSpace|'s [=coordinate system=].
```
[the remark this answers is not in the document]
toji: Done.

index.bs
NellWaliczek: Though the more i think about it, I'd kinda prefer if we define the algorithm but leave its contents blank.  That makes it really clear where Alex is expected to fill stuff in.

index.bs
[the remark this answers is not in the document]
NellWaliczek: Yeah I'm not sure calling the native origin an XRRigidTransform is correct... especially given that it does vary from frame to frame.

index.bs
```diff
@@ -670,40 +670,58 @@ An {{XRFrame}} represents a snapshot of the state of all of the tracked objects
...
 
-  1. If the {{XRFrame}}'s [=active=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
-  1. If the {{XRFrame}}'s [=animationFrame=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
-  1. Let |session| be the {{XRFrame}}'s {{XRFrame/session}} object.
-  1. If |referenceSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
-  1. If the [=viewer=]'s pose cannot be determined relative to |referenceSpace|, return <code>null</code>
-  1. Return a new {{XRViewerPose}} describing the [=viewer=]'s pose relative to the origin of |referenceSpace| at the timestamp of the {{XRFrame}}.
+  1. Let |frame| be the target {{XRFrame}}
+  1. Let |session| be |frame|'s {{XRFrame/session}} object.
+  1. If |frame|'s [=animationFrame=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
```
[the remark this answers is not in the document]
toji: No, because it only applies to `getViewerPose()`. This is the thing that prevents you from calling `getViewerPose()` in input events.

index.bs
Manishearth: This is kinda circling around the fact that there are two ways to talk of the transform from space A to space B.

You can:

 - Look at reference spaces as little bundles of coordinate grid lines, and talk of the transform used to take A's bundle of coordinate lines and turn it into those of B. This is _essentially_ what the current "natural origin" concept is describing. This is equivalent to the position and orientation of the coordinate grid of B _itself_ in A-space.
 - Look at the spaces as spaces an object lives in reference to, and talk of the transform used to take the coordinates of an object relative to A and produce the coordinates of an object relative to B. This is what @klausw is using

It's important to note that these two methods provide inverse transforms -- this is why Klaus' definition sounds "backwards".

This is also why I personally have been using the word "pose" a lot to underscore that I'm talking of the first method of looking at transforms -- talking about the position and orientation of the coordinate grid of B in A. While this does help get rid of ambiguity, it introduces a new ambiguity as "pose" is already used by the spec to mean something else. (I should stop using this, it slips through sometimes)

Anyway, I feel like this is the core mismatch in a bunch of discussions around transforms (it cropped up a lot in the originOffset discussion). We should probably pick one and be consistent about it (and define it thoroughly somewhere).

(In tensor algebra sometimes the concepts of covariance and contravariance can be used to formally distinguish between these two ways of looking at transforms, but I definitely don't think we should introduce that language in the spec.)

Manishearth: What we kind of need is a way to talk of a position and orientation _without_ calling it a "transform" since that term is overloaded. I prefer "pose" for this but that is already used by this spec.

index.bs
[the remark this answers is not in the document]
Manishearth: Ideally this is defined as some concept that is equivalent to a position and orientation, but is not called a "transform".

See also https://github.com/immersive-web/editor-collab/pull/36#discussion_r278361899

toji:
Okay, it's late and that probably means I'm doing something dumb, but I'm going to be unavailable for large parts of tomorrow and wanted to get something out for this before that so that I'm not leaving everyone hanging. From the comments above it seems pretty obvious to me that while we could likely stumble our way to a common ground it really just felt like we needed to establish a better baseline verbiage for talking about *what a space is*.

I did a bunch of reading about how similar concepts are talked about elsewhere and compared notes with how Klaus and Manish have been referring to different topics and finally feel like I'm getting a bit more comfortable with some of the language around these concepts. (Either that or the fact that it's almost midnight is imbuing me with unearned confidence.) I re-wrote basically the entire text to try and make these concepts more explicit and establish the lexicon around spaces better. I'd love some feedback on if that improves comprehension here or not.

Also note: Given that it seemed delicate enough to warrant it's own discussion I removed the transform multiplication algorithm from this PR. We can re-introduce it in a follow up if needed, but honestly I think the revised text is relatively clear without it, so maybe it's not as necessary as I assumed?

Also, one further question for Klaus specifically: You previously expressed to me that the term "transform" was potentially causing confusion here, especially since poses have a `.transform` attribute that makes them read somewhat awkwardly. A thought occurred to me while I was writing this: Would you find it clearer is `XRPose` extended `XRRigidTransform` instead of having one as a attribute? Then the code becomes:

```js
let pose = frame.getPose(spaceA, spaceB);
console.log(pose.position);
console.log(pose.orientation);
```
Rather than

```js
let pose = frame.getPose(spaceA, spaceB);
console.log(pose.transform.position);
console.log(pose.transform.orientation);
```
?

index.bs
```diff
@@ -926,9 +951,20 @@ Geometric Primitives {#geometricprimitives}
 Matrices {#matrices}
 --------
 
-WebXR provides various transforms in the form of <dfn lt="matrix">matrices</dfn>. WebXR matrices are always 4x4 and given as 16 element {{Float32Array}}s in column major order. They may be passed directly to WebGL's {{uniformMatrix4fv}} function, used to create an equivalent {{DOMMatrix}}, or used with a variety of third party math libraries.
+WebXR provides various transforms in the form of <dfn lt="matrix">matrices</dfn>. WebXR uses the WebGL conventions when communicating matrices, in which 4x4 matrices are given as 16 element {{Float32Array}}s with column major storage and column vectors. They may be passed directly to WebGL's {{uniformMatrix4fv}} function, used to create an equivalent {{DOMMatrix}}, or used with a variety of third party math libraries.
 
-Translations specified by WebXR matrices are always given in meters.
```
klausw: Did you intentionally remove this, or did it move somewhere else? I think specifying the unit is important.

klausw:
> I'd love some feedback on if that improves comprehension here or not.

I think it does help, I added some fairly small suggestions separately. IMHO, the core point to get across is that an XRSpace has two functions. It represents an origin as a real-world location, think of it as a set of XYZ basis arrows floating in the world somewhere. This origin lets you describe any point in space as coordinate values in the basis provided by that origin.

> Also note: Given that it seemed delicate enough to warrant it's own discussion I removed the transform multiplication algorithm from this PR. We can re-introduce it in a follow up if needed, but honestly I think the revised text is relatively clear without it, so maybe it's not as necessary as I assumed?

Manish said it was a useful concept for his followup PRs, so it seems as if we'd want to keep it, but this could indeed be a followup step.

> Also, one further question for Klaus specifically: You previously expressed to me that the term "transform" was potentially causing confusion here, especially since poses have a `.transform` attribute that makes them read somewhat awkwardly. A thought occurred to me while I was writing this: Would you find it clearer is `XRPose` extended `XRRigidTransform` instead of having one as a attribute?

Yes and no. I think it would help make code a bit more readable, but it doesn't directly address the underlying confusion, it just hides it from view. Short version, I don't see enough benefit in a restructuring or renaming to justify the churn in the spec, I think we can also solve this by having clear explanations in the spec, and the current proposal seems pretty close to that.

Longer version, I think the core conflict is that it may be unintuitive to say that a pose is-a transform.

A pose describes how something is placed in an underlying coordinate system, specifically that the position values are exactly the posed object's coordinates relative to the underlying space's origin.

A transform is a bit more complicated. I interpret it as a function that transforms an input value to an output value. XRRigidTransforms have a matrix member, and this matrix transforms coordinate values from the posed object's XRSpace to the underlying reference XRSpace, so in that sense it's very natural to interpret the XRRigidTransform as a transform from an object's space to the reference space. You can use this transform to answer the question "where is my object" by plugging in the object space origin coordinate values (all zeroes), and the result is the object's coordinates in the reference space, and this also works for non-origin points such as the geometry you want to draw for a controller model.

The potential conflict is that it sounds a bit backwards that an object's pose is a transform from object space to reference space. The opposing interpretation would be that a transform is an instruction how to move an object from base space origin to its current location. You still do that with the exact same matrix - plug in the old base space location at the base space origin (all zero values), and the result is the new coordinates in base space. Mathematically it's the same, but it's all staying in reference space, the object's XRSpace no longer appears. The problematic bit is that if you try to introduce the object's XRSpace into this picture, that's hard to do consistently, and some interpretations may end up expecting the inverse matrix to what we're currently using.

Overall though, I think this is getting a bit philosophical. As long as we clearly explain things, including sufficiently specific algorithms and examples to remove ambiguity, I think we're fine with the current terminology.

Manishearth:
> Also note: Given that it seemed delicate enough to warrant it's own discussion I removed the transform multiplication algorithm from this PR. We can re-introduce it in a follow up if needed, but honestly I think the revised text is relatively clear without it, so maybe it's not as necessary as I assumed?

In my followup PRs i found myself using it a bunch. It would be nice if we spec multiplication in one place, explicitly talking about pre/post multiplication and row/col vectors (etc) so we don't have to worry about clarity everywhere.

However for now we can leave it out if we want (some of my followup PRs will need to be updated)

> A thought occurred to me while I was writing this: Would you find it clearer is XRPose extended XRRigidTransform instead of having one as a attribute? Then the code becomes:

To me, yes, but then we should be very explicit in specifying what the underlying transform of a pose is, since the confusion will crop up there again (but then we only need to spec it out once!)

Overall to me this makes stuff nicer since we can just talk about "the pose of an object in this space", etc (e.g. when talking about the view offsets used by XRViewerPose)

index.bs
toji: The unit of measure is specified in the `XRRigidTransform` section, which felt more appropriate. This is old text left over from when we communicated EVERYTHING in terms of matrices, but it never felt quite accurate because it doesn't accurately describe things like projection matrices. (or, at least I don't *think* it does. Maybe I'm wrong on that count?)

index.bs
```diff
@@ -926,9 +951,20 @@ Geometric Primitives {#geometricprimitives}
...
+<div class="example">
+For a 16 element {{Float32Array}} such as this:
+<pre>[a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p]</pre>
```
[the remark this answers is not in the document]
toji: Yeah, I like that! I'll update the PR to use that example exactly as you typed it there, if you don't mind.

index.bs
```diff
@@ -682,48 +682,73 @@ The <dfn attribute for="XRFrame">session</dfn> attribute returns the {{XRSession
...
 Spaces {#spaces}
 ======
 
+A core feature of the WebXR Device API is the ability to provide spatial tracking. Spaces are the interface that enable applications to reason about how tracked entities are spatially related to the user's physical environment and each other. 
+
 XRSpace {#xrspace-interface}
 ------------------
 
-An {{XRSpace}} describes an entity that is tracked by the [=/XR device=]'s tracking systems. {{XRSpace}}s MAY NOT have a fixed spatial relationship to one another or to any given {{XRReferenceSpace}}. The transform between two {{XRSpace}} can be evaluated by calling an {{XRFrame}}'s {{XRFrame/getPose()}} method. The interface is intentionally opaque.
+An {{XRSpace}} represents a virtual coordinate system that corresponds to a physical location. Any time spatial data is requested from the API or given to the API, the {{XRSpace}} it is expressed in must be specified. The interface is intentionally opaque.
```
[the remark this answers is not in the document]
toji: 👍

index.bs
toji: In the process of doing my text refactor last night I found that, once rephrased, it was more intuitive to have this be one big helper function with slim method algorithms as you suggested.

toji:
Thank you everyone for the feedback! I'm feeling pretty good about where this is headed now because it seems like we're getting away from the "This is framing the concepts wrong.." discussions and into more "I would tweak this sentence like so..." comments. Thank you for your patience!

Latest commit applies Klaus' suggested text edits largely without change, and adds back in the transform multiplication bit since it seems less contentious and more helpful than I realized when I removed it last night. Sounds like there's some benefit to discussing the pose is-a/has-a transform question further, but that's also something that's best left to a separate issue, so I won't pursue it any further here.

This thread is long enough already that I could have easily missed some prior feedback in my last couple of revisions, so sorry if it seems like I ignored anything as that wasn't my intent!

Manishearth:
+1 on doing the pose-as-a-transform separately. I think the current state is not 100% clear, however I don't think we have to clarify it _now_, and this PR already does a bunch of work to make it clearer and I'd rather see this land first.

Let me know when you're done tweaking so I can update the offshoot PRs to fix conflicts (feel free to just merge them yourself)

index.bs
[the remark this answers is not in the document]
klausw: > Yeah, I like that! I'll update the PR to use that example exactly as you typed it there, if you don't mind.

Sure! Fair warning, I didn't try the formatting to see if it survives conversion correctly, especially the in-paragraph <pre>. Feel free to tweak as needed.

toji:
Sorry for the force-push noise, something got goofed up during a rebase and it took me a moment to notice and fix it.

So at this point it feels to me like this PR is in a happy place? Or at least a "better than we were before with no glaring new problems" place? As such, I'm going to move it over to the main /webxr repo as a PR soonish unless I hear panicked shouting. I think the editor collab repo has served it's purpose well of allowing quick iteration on a complex subject with a relatively small group, but I'd like to make sure we're doing most of our work in the more publicly visible venue.

Thanks again, everyone, for your support in getting these concepts sorted out!

Manishearth:
+1 on moving to the main repo. I may have some minor comments (need to read more thoroughly when on a computer again) but this looks pretty good as is.

Should I move my follow-up PRs too?

toji:
Yes, I think that once this PR is moved over it'd be a pretty sensible thing to move yours over too. If there's ongoing conversation happening in editor-collab, though, it's fine to wrap that up prior to moving.

index.bs
```diff
@@ -926,9 +951,25 @@ Geometric Primitives {#geometricprimitives}
...
+WebXR provides various transforms in the form of <dfn lt="matrix">matrices</dfn>. WebXR uses the WebGL conventions when communicating matrices, in which 4x4 matrices are given as 16 element {{Float32Array}}s with column major storage, and are applied to column vectors by premultiplying the matrix from the left. They may be passed directly to WebGL's {{uniformMatrix4fv}} function, used to create an equivalent {{DOMMatrix}}, or used with a variety of third party math libraries.
 
-Translations specified by WebXR matrices are always given in meters.
+<div class="example">
+Matrices returned from the WebXR Device API will be a 16 element {{Float32Array}} laid out like so:
+<pre>
+[a0, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15]
+</pre>
+
+Applying this matrix as a transform to a vector specified as a
```
klausw: optional: use "column vector" here.

klausw:
I think this is now in a good place, and I'm in favor of proceeding in the main repo. (I think it seems ready to merge as-is, but we should give others a chance to comment on it.)

index.bs
toji: Sounds good! Made this change prior to the move.

toji:
Thanks Klaus and Manish! Squashed this commit and moved it over to /webxr PR 609.

PR #37, opened 2019-04-23
Remove ability to dynamically modify originOffset

Manishearth:
Addresses point 3 of https://github.com/immersive-web/editor-collab/pull/36#pullrequestreview-229743358

This change actually makes it impossible to use originOffset -- maybe I misinterpreted that point?

r? @toji

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -772,10 +772,11 @@ enum XRStationaryReferenceSpaceSubtype {
 dictionary XRReferenceSpaceOptions {
   required XRReferenceSpaceType type;
   XRStationaryReferenceSpaceSubtype subtype;
+  XRRigidTransform originOffset;
 };
 
 [SecureContext, Exposed=Window] interface XRReferenceSpace : XRSpace {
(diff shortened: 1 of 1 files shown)


toji:
For that particular comment I believe Nell was referring to a change along the lines of PR 589 in the webxr repo, which makes the attribute readonly but also provides a way of setting it at reference space creation time. I've got an in-progress update to that particular PR that takes into account the current feedback that I can push soon.

Manishearth:
Ah, cool. I'll wait for that and then reflect it here.

Manishearth:
Updated anyway. To avoid changing too much at once I made `originOffset` an optional field on `XRReferenceSpaceOptions`, this could be changed to be more in line with https://github.com/immersive-web/webxr/pull/589 if necessary.

NellWaliczek:
I think this can be closed in favor of the PR @toji has in the main repo https://github.com/immersive-web/webxr/pull/589

PR #40, opened 2019-04-24
Provide alternate way to multiply matrices

Manishearth:
(I can write down the math proving why this works if folks need)

With everything in this spec being based on rigid transformations we don't actually have to require UAs to multiply-and-then-decompose matrices, since rigid transforms compose really nicely and efficiently without having to jump back and forth between matrix representations.

I'm not sure if this should be in the spec, but I'm wary about the current language requiring this to be a multiply-then-decompose since the answers there may slightly differ from the algorithm in this PR when it comes to floating point error. Not sure how much that matters, overall web specs don't seem to care too much about differences from float errors (hell, Blink and Gecko's CSS transform interpolation implementations are both completely different from the spec and each other)

Given all the confusion around matrices, speccing it this way (and including the matrix math in a note, i.e.  "multiplying `first` and `second` is equivalent to `first * second` in column-vector notation") might be one way to avoid having to ever talk about matrices. `inverse()` can be specced this way too.

Overall I'm on the fence about including this but I thought I'd put it up as an option.

r? @toji

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1022,13 +1022,25 @@ An {{XRRigidTransform}} with a {{XRRigidTransform/position}} of <code>{ x: 0, y:
 
 <div class="algorithm" data-algorithm="multiply transforms">
 
-To <dfn lt="multiply">multiply two {{XRRigidTransform}}s</dfn>, |first| and |second|, the UA MUST perform the following steps:
+To <dfn lt="multiply">multiply two {{XRRigidTransform}}s</dfn>, |first| and |second|, the UA MUST perform one of the following sets of steps:
+
+Matrix multiplication and decomposition:
 
   1. Let |result| be a new {{XRRigidTransform}} object.
   1. Set |result|'s {{XRRigidTransform/matrix}} to the result of multiplying |first|'s {{XRRigidTransform/matrix}} by |second|'s {{XRRigidTransform/matrix}}.
   1. Set |result|'s {{XRRigidTransform/orientation}} to the quaternion that describes the rotation indicated by the top left 3x3 sub-matrix of |result|'s {{XRRigidTransform/matrix}}.
   1. Set |result|'s {{XRRigidTransform/position}} to the vector given by the fourth column of |result|'s {{XRRigidTransform/matrix}}.
   1. Return |result|.
+
+Directly manipulating position and orientation:
+
+  1. Let |result| be a new {{XRRigidTransform}} object.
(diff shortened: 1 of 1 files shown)


Manishearth:
(This also is kinda independent of #36 and could be discussed on the main repo when #36 lands, feel free to close this)

index.bs
```diff
@@ -1022,14 +1022,27 @@ An {{XRRigidTransform}} with a {{XRRigidTransform/position}} of <code>{ x: 0, y:
...
   1. Let |result| be a new {{XRRigidTransform}} object.
   1. Set |result|'s {{XRRigidTransform/matrix}} to the result of multiplying |first|'s {{XRRigidTransform/matrix}} by |second|'s {{XRRigidTransform/matrix}}.
   1. Set |result|'s {{XRRigidTransform/orientation}} to the quaternion that describes the rotation indicated by the top left 3x3 sub-matrix of |result|'s {{XRRigidTransform/matrix}}.
   1. Set |result|'s {{XRRigidTransform/position}} to the vector given by the fourth column of |result|'s {{XRRigidTransform/matrix}}.
   1. Return |result|.
 
+Directly manipulating position and orientation:
+
+  1. Let |result| be a new {{XRRigidTransform}} object.
+  1. Let |t_1| be a vector that is the result of rotating |second|'s {{XRRigidTransform/position}} vector by the inverse of |first|'s {{XRRigidTransform/orientation}} quaternion.
```
klausw: Please use "multiplying" instead of "rotating" here. This is just a quaternion multiplication, and a vector rotation consists of two quaternion multiplications.

The single operation of rotating a vector by a quaternion `q` is calculated by first treating the vector as a quaternion `v={x, y, z, w=0}` and then calculating `v' = q * v * q^-1` using quaternion multiplication. You're doing exactly those two multiplications here and below for t_2, but that's just the separate steps of doing a single rotation.

If you compose the transforms using the matrix equivalents, the composed position is simply `B.matrix * A.position`, or (if decomposing the matrix into pure rotation and translation) `B.rotation_matrix * A.position + B.position`.

You can see this if you just multiply out `B.matrix * A.matrix` and look at the last column - that's exactly the A position vector [a12 a13 a14] rotated by B's 3x3 rotation submatrix with B's position vector [b12 b13 b14] added:

```
b0 b4 b8  b12 * a0 a4 a8  a12 = ? ? ? b0 * a12 + b4 * a13 + b8  * a14 + b12
b1 b5 b9  b13   a1 a5 a9  a13   ? ? ? b1 * a12 + b5 * a13 + b9  * a14 + b13
b2 b6 b10 b14   a2 a6 a10 a14   ? ? ? b2 * a12 + b6 * a13 + b10 * a14 + b14
0  0  0   1     0  0  0   1     ? ? ? 1
```

Manishearth: The reason I use "rotate" is that the notation is ambiguous: You can talk in terms of translation vectors, translation matrices, or translation quaternions, and rotation matrices, or quaternions. These all multiply differently, but the fundamental operation being done here is taking a translation and rotating it to produce another pure translation.

Here's the math, in pure matrix notation

image

Manishearth: > You're doing exactly those two multiplications here and below for t_2, but that's just the separate steps of doing a single rotation.

(I'd somewhat needlessly complicated the steps, I've simplified that)

Manishearth: The code should now reflect that the final translation is `B_rA_t + B_t`. I'd accidentally split a step where I didn't need to.

Manishearth: I made this step more redundant too. How does this look now?

PR #42, opened 2019-04-24
Explicitly specify how the views array is populated

Manishearth:
Address https://github.com/immersive-web/editor-collab/pull/36#discussion_r277831460

This does use the somewhat nebulous distinction between an `XRView` and a "view", and it doesn't really answer the question of where the views come from, instead just saying that the list of views depends on the underlying device. I'm not sure if you intended for this to be different.

r? @toji @NellWaliczek

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -389,6 +389,8 @@ enum XREnvironmentBlendMode {
 
 Each {{XRSession}} has a <dfn for="XRSession">mode</dfn>, which is one of the values of {{XRSessionMode}}.
 
+The {{XRSession/viewerSpace}} on an {{XRSession}} has a <dfn for="XRSession/viewerSpace">list of views</dfn>, which is a [=/list=] of [=view=]s corresponding to the views provided by the underlying device.
+
 <div class="algorithm" data-algorithm="initialize-session">
 
 To <dfn>initialize the session</dfn>, given |session| and |mode|, the user agent MUST run the following steps:
@@ -687,11 +689,19 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method 
   1. If |frame|'s [=animationFrame=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
   1. Let |pose| be a new {{XRViewerPose}} object.
   1. [=Populate the pose=] of |session|'s {{XRSession/viewerSpace}} in |referenceSpace| at the time represented by |frame| into |pose|.
+  1. If |pose| is <code>null</code> return <code>null</code>.
+  1. Let |xrviews| be an empty [=/list=].
+  1. For each [=view=] |view| in the [=XRSession/viewerSpace/list of views=] on the {{XRSession/viewerSpace}} of {{XRFrame/session}}, perform the following steps:
+      1. Let |xrview| be a new {{XRView}} object.
+      1. Initialize |xrview|'s {{XRView/eye}} to the appropriate eye the |view| is expected to render to. For views that have no such associated eye, this defaults to {{XREye/left}}.
+      1. Initialize |xrview|'s {{XRView/projectionMatrix}} to a [=matrix=] describing the projection to be used when rendering the [=view=], provided by the underlying XR system.
+      1. Let |offset| be an {{XRRigidTransform}} equal to the [=view offset=] of |view|
+      1. Set |xrview|'s {{XRView/transform}} property to the result of [=multiply transforms|multiplying=] the |offset| transform by the {{XRViewerPose}}'s {{XRPose/transform}}
+      1. [=list/Append=] |xrview| to |xrviews|
+  1. Set |pose|'ss {{XRViewerPose/views}} to |xrviews|
   1. Return |pose|.
 
 </div>
-
-ISSUE: Describe how the {{XRViewerPose/views}} are populated.
 
 <div class="algorithm unstable" data-algorithm="get-pose">
 
@@ -881,7 +891,11 @@ Views {#views}
 XRView {#xrview-interface}
 ------
 
-An {{XRView}} describes a single <dfn>view</dfn> into an XR scene. Each [=view=] corresponds to a display or portion of a display used by an XR device to present imagery to the user. They are used to retrieve all the information necessary to render content that is well aligned to the [=view=]'s physical output properties, including the field of view, eye offset, and other optical properties. [=Views=] may cover overlapping regions of the user's vision. No guarantee is made about the number of [=views=] any XR device uses or their order, nor is the number of [=views=] required to be constant for the duration of an {{XRSession}}.
+An {{XRView}} describes a single <dfn>view</dfn> into an XR scene for a given frame.
+
+Each [=view=] corresponds to a display or portion of a display used by an XR device to present imagery to the user. They are used to retrieve all the information necessary to render content that is well aligned to the [=view=]'s physical output properties, including the field of view, eye offset, and other optical properties. [=Views=] may cover overlapping regions of the user's vision. No guarantee is made about the number of [=views=] any XR device uses or their order, nor is the number of [=views=] required to be constant for the duration of an {{XRSession}}.
+
+A [=view=] has an associated internal <dfn>view offset</dfn>, which is an {{XRRigidTransform}} describing the position and orientation of the [=view=] in the {{XRSession/viewerSpace}}'s [=coordinate system=].
 
 NOTE: Many HMDs will request that content render two [=views=], one for the left eye and one for the right, while most magic window devices will only request one [=view=], but applications should never assume a specific view configuration. For example: A magic window device may request two views if it is capable of stereo output, but may revert to requesting a single view for performance reasons if the stereo output mode is turned off. Similarly, HMDs may request more than two views to facilitate a wide field of view or displays of different pixel density.
 


index.bs
```diff
@@ -1122,6 +1128,22 @@ The <dfn attribute for="XRViewerPose">views</dfn> array is a sequence of {{XRVie
 
 NOTE: The {{XRViewerPose}}'s {{XRPose/transform}} can be used to position graphical representations of the [=viewer=] for spectator views of the scene or multi-user interaction.
 
+<div class="algorithm" data-algorithm="initialize views">
```
NellWaliczek: Depending on how much we can clean up the getViewerPose algorithm, it may make more sense just to put this in there rather than have a separate algorithm since it's only called in one place.

index.bs
```diff
@@ -1122,6 +1128,22 @@ The <dfn attribute for="XRViewerPose">views</dfn> array is a sequence of {{XRVie
...
+
+To <dfn for="XRViewerPose">initialize the views</dfn> on an {{XRViewerPose}} given an {{XRSession}} |session| the UA MUST perform the following steps:
+
+  1. Let |views| be an empty [=/list=].
```
NellWaliczek: call this "xrviews" if the thing going into it is an "xrview"

index.bs
```diff
@@ -1122,6 +1128,22 @@ The <dfn attribute for="XRViewerPose">views</dfn> array is a sequence of {{XRVie
...
+  1. For each [=view=] |view| in |session|'s [=XRSession/list of views=], perform the following steps:
+      1. Let |xrview| be a new {{XRView}} object.
+      1. Initialize |xrview|s {{XRView/eye}} to the appropriate eye the |view| is expected to render to. For views that have no such associated eye, this defaults to {{XREye/left}}.
```
NellWaliczek: This should default to "".

index.bs
```diff
@@ -1122,6 +1128,22 @@ The <dfn attribute for="XRViewerPose">views</dfn> array is a sequence of {{XRVie
...
+      1. Initialize |xrview|s {{XRView/projectionMatrix}} to a [=matrix=] describing the projection to be used when rendering the [=view=]
```
NellWaliczek: missing apostrophe

NellWaliczek: Also, let's call out that this info should be provided by the underlying XR System.

index.bs
```diff
@@ -1122,6 +1128,22 @@ The <dfn attribute for="XRViewerPose">views</dfn> array is a sequence of {{XRVie
...
+      1. Let |offset| be an {{XRRigidTransform}} with {{XRRigidTransform/position}} equal to the [=view offset=] of |view|
```
NellWaliczek: This needs to support including orientation as well.  Frankly this offset should be used as-is from the underlying XR system.

index.bs
```diff
@@ -890,7 +892,11 @@ Views {#views}
...
+An {{XRView}} describes a single <dfn>view</dfn> into an XR scene for a given frame.
+
+Each [=view=] corresponds to a display or portion of a display used by an XR device to present imagery to the user. They are used to retrieve all the information necessary to render content that is well aligned to the [=view=]'s physical output properties, including the field of view, eye offset, and other optical properties. [=Views=] may cover overlapping regions of the user's vision. No guarantee is made about the number of [=views=] any XR device uses or their order, nor is the number of [=views=] required to be constant for the duration of an {{XRSession}}.
+
+A [=view=] has an associated internal <dfn>view offset</dfn>, which is the pose of the eye relative to the [=viewer=].
```
NellWaliczek: > pose of the 

This should explicitly not be a "pose", but rather an `XRRigidTransform`

NellWaliczek: > relative to the [=viewer=]

Perhaps we want to say relative to the `viewerSpace`

index.bs
```diff
@@ -1122,6 +1128,22 @@ The <dfn attribute for="XRViewerPose">views</dfn> array is a sequence of {{XRVie
...
+
+To <dfn for="XRViewerPose">initialize the views</dfn> on an {{XRViewerPose}} given an {{XRSession}} |session| the UA MUST perform the following steps:
+
+  1. Let |views| be an empty [=/list=].
+  1. For each [=view=] |view| in |session|'s [=XRSession/list of views=], perform the following steps:
+      1. Let |xrview| be a new {{XRView}} object.
+      1. Initialize |xrview|s {{XRView/eye}} to the appropriate eye the |view| is expected to render to. For views that have no such associated eye, this defaults to {{XREye/left}}.
+      1. Initialize |xrview|s {{XRView/projectionMatrix}} to a [=matrix=] describing the projection to be used when rendering the [=view=]
+      1. Let |offset| be an {{XRRigidTransform}} with {{XRRigidTransform/position}} equal to the [=view offset=] of |view|
+      1. Set |xrview|'s {{XRView/transform}} property to the result of [=multiply=]ing |offset| by the {{XRViewerPose}}'s {{XRPose/transform}}
```
NellWaliczek: We should be explicit about which multiply algorithm we mean.  In this case, we're talking about multiplying two transforms.

index.bs
```diff
@@ -1122,6 +1128,22 @@ The <dfn attribute for="XRViewerPose">views</dfn> array is a sequence of {{XRVie
...
+      1. [=list/Append=] |xrview| to |views|
+  1. Set {{XRViewerPose}};ss {{XRViewerPose/views}} to |views|
```
NellWaliczek: > ;ss

typo

index.bs
```diff
@@ -389,6 +389,9 @@ enum XREnvironmentBlendMode {
 
 Each {{XRSession}} has a <dfn for="XRSession">mode</dfn>, which is one of the values of {{XRSessionMode}}.
 
+
+Each {{XRSession}} has a <dfn for="XRSession">list of views</dfn>, which is a [=/list=] of [=view=]s corresponding to the views provided by the underlying device.
```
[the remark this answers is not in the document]
Manishearth: Ooh, I like that.

index.bs
Manishearth: Yeah, I'm trying to keep it separate for now since there are other PRs messing with that algorithm too. Perhaps we should revisit this once everything has landed?

index.bs
Manishearth: It can't, the enum has no such variant, and the spec already defaults this to left: https://immersive-web.github.io/webxr/#dom-xrview-eye

index.bs
Manishearth: Ah, missed that.

index.bs
Manishearth: Ah, in that case we hit on the ambiguity when talking about transforms again. I'll try to clarify.

Manishearth: Done, unsure if this is clear enough.

index.bs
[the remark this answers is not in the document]
Manishearth: Done.

index.bs
Manishearth: We only have one in the spec right now, yes? Added some language to clarify.

Manishearth:
Addressed

index.bs
```diff
@@ -856,7 +891,11 @@ Views {#views}
...
+An {{XRView}} describes a single <dfn>view</dfn> into an XR scene for a given frame.
+
+Each [=view=] corresponds to a display or portion of a display used by an XR device to present imagery to the user. They are used to retrieve all the information necessary to render content that is well aligned to the [=view=]'s physical output properties, including the field of view, eye offset, and other optical properties. [=Views=] may cover overlapping regions of the user's vision. No guarantee is made about the number of [=views=] any XR device uses or their order, nor is the number of [=views=] required to be constant for the duration of an {{XRSession}}.
+
+A [=view=] has an associated internal <dfn>view offset</dfn>, which is the {{XRRigidTransform}} describing the position and orientation of the [=view=] relative to the {{XRSession/viewerSpace}}.
```
NellWaliczek: Can we update this sentence to follow the patterns in https://github.com/immersive-web/webxr/pull/589  Something like 

> the {{XRRigidTransform}} describing the position and orientation of the [=view=] in {{XRSession/viewerSpace}}'s [=coordinate system=].

index.bs
```diff
@@ -986,6 +1025,18 @@ The <dfn attribute for="XRRigidTransform">inverse</dfn> attribute returns a {{XR
 
 An {{XRRigidTransform}} with a {{XRRigidTransform/position}} of <code>{ x: 0, y: 0, z: 0 w: 1 }</code> and an {{XRRigidTransform/orientation}} of <code>{ x: 0, y: 0, z: 0, w: 1 }</code> is known as an <dfn>identity transform</dfn>.
 
+<div class="algorithm" data-algorithm="multiply transforms">
```
NellWaliczek: This is no longer needed, yes?

index.bs
```diff
@@ -1076,6 +1127,22 @@ The <dfn attribute for="XRViewerPose">views</dfn> array is a sequence of {{XRVie
 
 NOTE: The {{XRViewerPose}}'s {{XRPose/transform}} can be used to position graphical representations of the [=viewer=] for spectator views of the scene or multi-user interaction.
 
+<div class="algorithm" data-algorithm="initialize views">
+
+To <dfn for="XRViewerPose">initialize the views</dfn> on an {{XRViewerPose}} given an {{XRSession}} |session| the UA MUST perform the following steps:
+
+  1. Let |xrviews| be an empty [=/list=].
+  1. For each [=view=] |view| in the [=XRSession/list of views=] on the {{XRSession/viewerSpace}} of |session|, perform the following steps:
```
NellWaliczek: >[=XRSession/list of views=]

This is the viewerSpace now, yes?

Manishearth: Yeah I just wasn't sure if `XRSession/viewerSpace/list of views` would work. I'll try it and update.

Manishearth:
Updated.

index.bs
```diff
@@ -687,12 +689,21 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method
   1. If |frame|'s [=animationFrame=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
   1. Let |pose| be a new {{XRViewerPose}} object.
   1. [=Populate the pose=] of |session|'s {{XRSession/viewerSpace}} in |referenceSpace| at the time represented by |frame| into |pose|.
+  1. Set |pose|'s {{XRPose/transform}} to the [=transform from=] |session|'s {{XRSession/viewerSpace}} to |referenceSpace| at the time represented by |frame|.
```
NellWaliczek: I think this line is no longer needed due to the line above it?

index.bs
```diff
@@ -687,12 +689,21 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method
...
+  1. If |pose|'s {{XRPose/transform}} is <code>null</code> return <code>null</code>.
+  1. Let |xrviews| be an empty [=/list=].
+  1. For each [=view=] |view| in the [=XRSession/viewerSpace/list of views=] on the {{XRSession/viewerSpace}} of {{XRFrame/session}}, perform the following steps:
+      1. Let |xrview| be a new {{XRView}} object.
+      1. Initialize |xrview|'s {{XRView/eye}} to the appropriate eye the |view| is expected to render to. For views that have no such associated eye, this defaults to {{XREye/left}}.
+      1. Initialize |xrview|'s {{XRView/projectionMatrix}} to a [=matrix=] describing the projection to be used when rendering the [=view=], provided by the underlying XR system.
+      1. Let |offset| be an {{XRRigidTransform}} equal to the [=view offset=] of |view|
+      1. Set |xrview|'s {{XRView/transform}} property to the result of [=multiply=]ing the |offset| transform by the {{XRViewerPose}}'s {{XRPose/transform}}
+      1. [=list/Append=] |xrview| to |xrviews|
+  1. Set {{XRViewerPose}} {{XRViewerPose/views}} to |xrviews|
```
NellWaliczek: Shouldn't this set |pose|?

index.bs
```diff
@@ -687,12 +689,21 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method
...
+  1. Set |pose|'s {{XRPose/transform}} to the [=transform from=] |session|'s {{XRSession/viewerSpace}} to |referenceSpace| at the time represented by |frame|.
+  1. If |pose|'s {{XRPose/transform}} is <code>null</code> return <code>null</code>.
+  1. Let |xrviews| be an empty [=/list=].
+  1. For each [=view=] |view| in the [=XRSession/viewerSpace/list of views=] on the {{XRSession/viewerSpace}} of {{XRFrame/session}}, perform the following steps:
+      1. Let |xrview| be a new {{XRView}} object.
+      1. Initialize |xrview|'s {{XRView/eye}} to the appropriate eye the |view| is expected to render to. For views that have no such associated eye, this defaults to {{XREye/left}}.
+      1. Initialize |xrview|'s {{XRView/projectionMatrix}} to a [=matrix=] describing the projection to be used when rendering the [=view=], provided by the underlying XR system.
+      1. Let |offset| be an {{XRRigidTransform}} equal to the [=view offset=] of |view|
+      1. Set |xrview|'s {{XRView/transform}} property to the result of [=multiply=]ing the |offset| transform by the {{XRViewerPose}}'s {{XRPose/transform}}
```
NellWaliczek: >[=multiply=]

I don't think this is the algorithm name anymore?

NellWaliczek: Yeah, double checked and it's "multiply transforms" now.

Manishearth:
Now at https://github.com/immersive-web/webxr/pull/610

PR #43, opened 2019-06-12
Adding tracking loss logic to the spec

toji:
So this feels anticlimactic.

I've spent a ton of time stressing over how to capture everything in Alex's tracking loss explainer PR, and ended up with huge chunks of non-normative text that echoed what he put in the explainer but which I wasn't sure really provided any clarity to the spec specifically. There's a lot of what Alex describes that's best practices, or examples of scenarios in which hardware may choose to apply some bit of logic, but not a whole lot that defines actual API behavior.

I finally decided to try stripping away the non-normative bits and just see what it would look like if I only wrote up the actual functional changes and, well... it's actually pretty tiny. To the point where I'm concerned that I'm missing something. (Paging @thetuvix for a sanity check!) But then again the point of this was never that is was complex, just that we wanted to ensure developers were enabled to do the right thing in every situation.

So my thoughts are: Let's make sure this is actually capturing the behavioral changes in the way we want, then push it and the explainer changes through together. After that we likely do want to spend more time figuring out the right way to surface some of the non-normative examples and best practices in the spec itself (especially since we're already planning on expanding descriptive text elsewhere), but that's not critical for the June milestone.


index.bs
```diff
@@ -805,16 +805,26 @@ The [=effective origin=] of an {{XRSpace}} can only be observed in the coordinat
...
-  1. If |baseSpace|'s pose cannot be determined relative to |space| at the time represented by |frame|, set |pose| to <code>null</code>.
   1. Let |transform| be |pose|'s {{XRPose/transform}}.
-  1. Set |transform|'s {{XRRigidTransform/position}} to the location of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
-  1. Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+  1. Set |pose| based on the following:
+    <dl class="switch">
+      <dt> If |space|'s pose relative to |baseSpace| at the time represented by |frame| can be determined.
+      <dd> Set |transform|'s {{XRRigidTransform/position}} to the location of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+      <dd> Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+      <dd> If |transform|'s {{XRRigidTransform/position}} was computed based on |space|'s orientation, such as via a neck or elbow model, set |pose|'s {{XRPose/emulatedPosition}} boolean to <code>true</code>.
```
NellWaliczek: This seems a bit.... wishy washy?  Is there any way to be more specific?  Like... referring to the 3DOF-ness or 6DOF-ness of the underlying tracking tech?  Maybe...
> <dd>Set |pose|'s {{XRPose/emulatedPosition}} to true If the either |baseSpace|'s or |space|'s underlying tracking technology is only able to track three degrees of freedom, but applies a computed offset to simulate any amount of six degree of freedom motion.

index.bs
```diff
@@ -805,16 +805,26 @@ The [=effective origin=] of an {{XRSpace}} can only be observed in the coordinat
...
   1. 1. If |frame|'s [=active=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
   1. Let |session| be |frame|'s {{XRFrame/session}} object.
   1. If |space|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
   1. If |baseSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
-  1. If |baseSpace|'s pose cannot be determined relative to |space| at the time represented by |frame|, set |pose| to <code>null</code>.
   1. Let |transform| be |pose|'s {{XRPose/transform}}.
-  1. Set |transform|'s {{XRRigidTransform/position}} to the location of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
-  1. Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+  1. Set |pose| based on the following:
+    <dl class="switch">
```
NellWaliczek: What should |emulatedPosition| be set to if using a local-floor with an emulated floor level?

index.bs
```diff
@@ -805,16 +805,26 @@ The [=effective origin=] of an {{XRSpace}} can only be observed in the coordinat
...
+      <dt> If |space|'s pose relative to |baseSpace| at the time represented by |frame| can be determined.
```
NellWaliczek: This phrasing seems a bit too vague... I'm a bit worried about a disconnect in behavior if underlying tracking tech reports this differently.  For example, one reports it cannot determine a pose and another determines it can but sets it's own version of 'emulatedPosition'.  I wonder if there's a way we can be more crisp?

index.bs
```diff
@@ -805,16 +805,26 @@ The [=effective origin=] of an {{XRSpace}} can only be observed in the coordinat
...
 
   1. 1. If |frame|'s [=active=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
   1. Let |session| be |frame|'s {{XRFrame/session}} object.
   1. If |space|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
   1. If |baseSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
-  1. If |baseSpace|'s pose cannot be determined relative to |space| at the time represented by |frame|, set |pose| to <code>null</code>.
   1. Let |transform| be |pose|'s {{XRPose/transform}}.
-  1. Set |transform|'s {{XRRigidTransform/position}} to the location of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
-  1. Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+  1. Set |pose| based on the following:
```
NellWaliczek: Ok, with the shift to modularization... how do we intend to handle if subtypes of XRSpace don't follow this pattern?  For example, perhaps we'll decide anchors should always report null instead of an emulated pose?

index.bs
```diff
@@ -1931,7 +1941,7 @@ A user agent MUST dispatch a <dfn event for="XRSession">selectend</dfn> event on
...
+A user agent MUST dispatch a <dfn event for="XRReferenceSpace">reset</dfn> event on an {{XRReferenceSpace}} when discontinuities of the [=native origin=] or [=effective origin=] occur, i.e. there are significant changes in the origin’s position or orientation relative to the user’s environment. (For example: After user recalibration of their XR device or if the XR device automatically shifts its origin after losing and regaining tracking.) A {{reset}} event MUST also be dispatched when the {{boundsGeometry}} changes for a {{XRBoundedReferenceSpace}}. {{reset}} events SHOULD NOT be dispatched if the [=viewer=]'s pose experiences discontinuities but the {{XRReferenceSpace}}'s origin remains stable. The event MUST be of type {{XRReferenceSpaceEvent}}, and MUST be dispatched prior to the execution of any [=XR animation frame=]s that make use of the new origin.
```
NellWaliczek: > discontinuities of the [=native origin=] or [=effective origin=] occur, i.e. there are significant changes in the origin’s position or orientation relative to the user’s environment.

Ooooh, maybe we can define this better now!  Like... we can divide it into two categories: abrupt changes like recalibration and when emulatedPose switches from reporting false to true... wdyt?

index.bs
toji: I see what you're getting at, and I'll try updating it. We want to make sure we also include 6DoF hardware that has temporarily fallen back to 3DoF behavior too, however.

index.bs
toji: Ewe. I'm gonna say.... that shouldn't matter? Whether or not the _origin_ of the space is emulated isn't the question this field is trying to answer. It's whether or not the pose relative to that origin is emulated. (At least, that sounded good in my head when I made it up just now.)

index.bs
toji: What about something like:

> Set |pose|'s {{XRPose/emulatedPosition}} to <code>true</code> if the relationship between |space| and |baseSpace| is only able to be evaluated via [=3DoF=] tracking, but a computed offset was applied to simulate any amount of [=6DoF=] motion.

(Yay for having 3DoF and 6DoF as [defined terms in the spec](http://localhost:8080/webxr/#3dof) already!)

index.bs
toji: This is tough, especially because I don't have a clear idea in mind about what circumstances this might apply to, but I've tweaked the line about setting the `emulatedPosition` flag to include:

> ... or if the underlying tracking system otherwise indicates that the position was an emulated value.

index.bs
toji: I'm willing to retroactively change the core if we need to introduce language hooks that enable any given module. Discounting that, however, I would anticipate that the language in the module would be something along the lines of "If <condition is met> |space| MUST NOT report that a pose relative to |baseSpace| can be determined."

index.bs
toji: I'm honestly not sure what this would entail beyond what I've got there. (I did add a non-normative note indicating how to correlate positional jumps with `emulatedPosition` to detect tracking loss/resume.)

toji:
Made some updates, but I'm really not sure how well they address your comments. Wanted to ask about it on the call today but we ran out of time.

index.bs
NellWaliczek: Seems reasonable, but we should state that explicitly as a non- normative note perhaps.

index.bs
NellWaliczek: Not sure I'm following you on this?  Once we've hit Rec, isn't changing the core a huge hassle?

Alternatively, are you suggesting the module overrides the core behavior?

index.bs
```diff
@@ -837,16 +837,26 @@ The [=effective origin=] of an {{XRSpace}} can only be observed in the coordinat
...
   1. Let |transform| be |pose|'s {{XRPose/transform}}.
-  1. Set |transform|'s {{XRRigidTransform/position}} to the location of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
-  1. Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+  1. Set |pose| based on the following:
+    <dl class="switch">
+      <dt> If |space|'s pose relative to |baseSpace| at the time represented by |frame| can be determined.
+        <dd> Set |transform|'s {{XRRigidTransform/position}} to the location of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |pose|'s {{XRPose/emulatedPosition}} to <code>true</code> if the relationship between |space| and |baseSpace| is only able to be evaluated via [=3DoF=] tracking but a computed offset was applied to simulate any amount of [=6DoF=] motion, or if the underlying tracking system otherwise indicates that the position was an emulated value.
```
NellWaliczek: >Set |pose|'s {{XRPose/emulatedPosition}} to <code>true</code> if the relationship between |space| and |baseSpace| is only able to be evaluated via [=3DoF=] tracking but a computed offset was applied to simulate any amount of [=6DoF=] motion, or if the underlying tracking system otherwise indicates that the position was an emulated value.

Should emulatedPosition be true even if no additional 6DOF parts are added?  Or no?  My inclination is yes, but i'm not sure if that matches Alex's intention...

index.bs
```diff
@@ -1575,7 +1585,7 @@ dictionary XRWebGLLayerInit {
   double framebufferScaleFactor = 1.0;
 };
 
-[SecureContext, Exposed=Window, Constructor(XRSession session,
+[SecureContext, Exposed=Window, Constructor(XRSession session,`
```
NellWaliczek: Is that character necessary?

index.bs
NellWaliczek: I'm not 100% Alex will be on board with this language, but I'm happy to defer to him on it :)

thetuvix:
Thanks so much for pulling this all together!  I got sick on Friday, so I wasn't able to look at this last week - I should be able to get you some thoughts by end of day.

index.bs
toji: Nope.

index.bs
toji: I guess I was mentally understimating the effort required to change core after CR, but for the modules I think it should be fine to allow them to override core behavior.

It's not a perfect analog, but in OpenGL land many of the extensions contain text to the effect of "Change section 2.3.4 paragraph 1 to read: <lots o' text>." I don't advocate for THAT level of granularity, but having modules that occasionally say things like "Add <value> to the list of accepted values" or "Use this variant of the algorithm instead" doesn't seem unreasonable.

index.bs
toji: Done

index.bs
toji: Not sure I grok what you're asking here, but the non-normative note I added at the end of thie algorithm may clear it up anyway?

toji:
Addressed the last of Nell's comments. Alex: After you have a chance to take a look if you feel like it's ready to go then Nell and I feel like this change is compact enough that we could just make it part of your webxr/559 change. (If you'd prefer to keep them separate that's fine too, we'll just want to land them and roughly the same time.)

index.bs
thetuvix: I think Nell's question is asking about a 3DoF device that doesn't even bother with a neck model - full speed ahead with just orientation.  If that device returns you a position component of (0, 0, 0) 100% of the time, would we set `emulatedPosition` to true?

My take is yes, since that is a degenerate case of most real-world devices where at least some attempt will be made to give out a live position.  I believe developers will treat `emulatedPosition` as `lowQualityPosition` or `3dofPosition`, treating `false` as a sign of full positional tracking.  We shouldn't have a donut hole here where great tracking gets you false, medium tracking gets you true, and then the worst tracking gets you false again.

thetuvix: My [spatial tracking explainer updates](https://github.com/immersive-web/webxr/pull/559/files#diff-f02c24f867e61b8e58dc8f2c228f476cR221) phrase this in the opposite way, defining what false means instead:
> The `emulatedPosition` attribute of `XRPose` indicates that the position component of the retrieved pose matrix does not represent an actively tracked position.

That might be a useful approach.

This could also be a smell that we have the wrong name for `emulatedPosition` if it's not always about emulation. 
 OpenXR calls the equivalent bit `XR_SPACE_LOCATION_POSITION_TRACKED_BIT` 😇

toji: Ah! That makes sense, and I completely agree. Now the tricky part: how to phrase that? Hm... Something like this perhaps?

> Set |pose|'s {{XRPose/emulatedPosition}} to true if the relationship between |space| and |baseSpace| is unable to to be fully evaluated in [=6DoF=]. This includes [=3DoF=] tracking where a computed offset was applied to simulate any amount of [=6DoF=] motion, or if the underlying tracking system otherwise indicates that the position was an emulated value.

index.bs
```diff
@@ -844,16 +844,28 @@ The [=effective origin=] of an {{XRSpace}} can only be observed in the coordinat
...
   1. Let |transform| be |pose|'s {{XRPose/transform}}.
-  1. Set |transform|'s {{XRRigidTransform/position}} to the location of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
-  1. Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+  1. Set |pose| based on the following:
+    <dl class="switch">
+      <dt> If |space|'s pose relative to |baseSpace| at the time represented by |frame| can be determined.
+        <dd> Set |transform|'s {{XRRigidTransform/position}} to the location of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |pose|'s {{XRPose/emulatedPosition}} to <code>true</code> if the relationship between |space| and |baseSpace| is only able to be evaluated via [=3DoF=] tracking but a computed offset was applied to simulate any amount of [=6DoF=] motion, or if the underlying tracking system otherwise indicates that the position was an emulated value.
```
thetuvix: There is a broader concern with this line that perhaps speaks to @toji's general angst about whether he'd captured enough of my intention from the explainer, or left too much to be non-normative :smile:

As phrased here, this sentence tells a UA when to set `emulatedPosition` to `true` or `false`, based on whether it happened to emulate any part of the `position` component of the transform.  However, [my explainer went a little further](https://github.com/immersive-web/webxr/pull/559/files#diff-f02c24f867e61b8e58dc8f2c228f476cR221), imploring UA's that they "should" continue to provide positions even during tracking loss:
> However, once a pose is initially established for a space (e.g. a `viewer` reference space or an input source's `gripSpace`), pose matrices should continue to be provided even during tracking loss.

Apps will be targeting the WebXR API through some number of UAs that then build on a broader number of native XR platforms.  The goal with my tracking loss updates was to make it the job of the UA to absorb that native complexity - even if the underlying XR platform happens to just toss the UA null positions when positional tracking is lost, it's then the job of the UA to remember the last position and return it to the app with `emulatedPosition` set to true, continuing to update the orientation if possible.

I think the failure state for us here is if WebXR apps are subject to the whims of how 10 different underlying native XR APIs happen to have designed their tracking loss semantics years ago, when it's within reach for the UAs to just agree on some commonly-implementable semantics (which is what we did in https://github.com/immersive-web/webxr/issues/243.

thetuvix: Put a different way, I'd debated in my explainer updates whether UA-targeted "should"s like the one above should instead have been "must"s.  Here, some of the critical "should"s have become "non-normative", which I believe will result in unnecessary divergence among UAs and XR platforms.

thetuvix: See, now I kept reading the next few lines and I see you covered that explicitly 😅

index.bs
thetuvix: Better!  Although I think the second sentence still sort of makes it sound like it matters whether a computed offset was applied.

Perhaps:
> Set |pose|'s {{XRPose/emulatedPosition}} to true if the relationship between |space| and |baseSpace| is unable to be fully evaluated in [=6DoF=]. This includes [=3DoF=] tracking (whether or not a computed offset such as a neck model is applied to simulate any amount of [=6DoF=] motion), or if the underlying tracking system otherwise indicates that the position was an emulated value.

index.bs
thetuvix: OK - reading through this some more, I believe there is some ambiguity here.

The parent bullet here talks about whether `space`'s pose "can be determined".  However, it's not clear whether we're talking about a 6DoF pose or a 3DoF pose.  Depending on how the reader interprets that, they'll end up down one garden path or another.

For example, if they assume that the bullet is talking only about full 6DoF poses, then during positional tracking loss when only orientation is available, they'll go into the second bullet.  However, then they'll miss all your coverage about computed tracking offsets.

If they notice that stuff, they'll probably assume they were meant to read it as counting 3DoF poses as "poses" too.  However, then they'll only take it upon themselves to reuse "last known" positions when the underlying XR platform gives them zilch - it's then undefined what to do if the underlying platform gives you orientation with a null position.

I think the solution here is to have the top-level bullets reason about "**position**"" rather than "pose", specifically "actively-tracked position" or such.  We then have a few clear cases:
* If the underlying platform is giving you actively-tracked position, set the position and orientation appropriately and then set `emulatedPosition` to false.
* If the underlying platform is **not** giving you actively-tracked position, use the best emulated position available from the platform.  If the platform is giving you no position at all, go ahead and use the last-known position.  Pair that position with whatever orientation you're getting from the platform (and any neck model offset, if it's delivered separately from that missing position - *cough* Windows MR *cough*).  If you're getting no orientation either, go ahead and use the last-known orientation too.
* If you haven't tracked at all yet, set the whole pose to null.

index.bs
```diff
@@ -1956,7 +1966,9 @@ A user agent MUST dispatch a <dfn event for="XRSession">selectend</dfn> event on
...
+A user agent MUST dispatch a <dfn event for="XRReferenceSpace">reset</dfn> event on an {{XRReferenceSpace}} when discontinuities of the [=native origin=] or [=effective origin=] occur, i.e. there are significant changes in the origin’s position or orientation relative to the user’s environment. (For example: After user recalibration of their XR device or if the XR device automatically shifts its origin after losing and regaining tracking.) A {{reset}} event MUST also be dispatched when the {{boundsGeometry}} changes for a {{XRBoundedReferenceSpace}}. {{reset}} events SHOULD NOT be dispatched if the [=viewer=]'s pose experiences discontinuities but the {{XRReferenceSpace}}'s origin remains stable, such as when the [=viewer=] loses and regains tracking. The event MUST be of type {{XRReferenceSpaceEvent}}, and MUST be dispatched prior to the execution of any [=XR animation frame=]s that make use of the new origin.
```
thetuvix: The "such as" here for UAs is generally correct...

> {{reset}} events SHOULD NOT be dispatched if the [=viewer=]'s pose experiences discontinuities but the {{XRReferenceSpace}}'s origin remains stable, such as when the [=viewer=] loses and regains tracking.

...but it's missing the exception mentioned [here in the explainer](https://github.com/immersive-web/webxr/pull/559/files#diff-f02c24f867e61b8e58dc8f2c228f476cR242).

We likely just need to condition the "such as" to refer to regaining tracking "without a shiff in the origin's physical mapping" or such.  That would be more consistent anyway with the parenthetical earlier in the paragraph that talks about how the origin may shift "after losing and regaining tracking".

toji:
Okay, took another pass at things based on Alex's feedback (thank you!) I'm not sure if this'll be "the one", but I feel like we're definitely closer now!

For the record, I'm not opposed to a more OpenXR-ish name like `positionTracked` for the `emulatedPosition` bit, as it likely does reflect the usage better. I'm not inclined to make that change here, though, and I'm doubtful that the name is so egregious that we want to push to rename prior to VR complete.

index.bs
toji: I think I captured this far better in the latest revision. PTAL!

toji:
@thetuvix: Any further feedback on this? I'd like to get this and your explainer changes merged ASAP if we're happy with the current state of things.

toji:
FYI: I took a moment to rebase Alex's original PR onto the current master and then rebase my spec change on top of it. It's in this repo's [full-tracking-pr branch](https://github.com/immersive-web/editor-collab/compare/full-tracking-pr) if you're interested.

index.bs
```diff
@@ -844,16 +844,28 @@ The [=effective origin=] of an {{XRSpace}} can only be observed in the coordinat
...
   1. If |space|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
   1. If |baseSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
-  1. If |baseSpace|'s pose cannot be determined relative to |space| at the time represented by |frame|, set |pose| to <code>null</code>.
   1. Let |transform| be |pose|'s {{XRPose/transform}}.
-  1. Set |transform|'s {{XRRigidTransform/position}} to the location of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
-  1. Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+  1. Set |pose| based on the following:
+    <dl class="switch">
+      <dt> If |space|'s pose relative to |baseSpace| at the time represented by |frame| can be determined.
+        <dd> Set |transform|'s {{XRRigidTransform/position}} to the location of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
```
thetuvix: Nit: "position" rather than "location" here - we are otherwise consistent on using "position", as "location" may imply rotation as well.

index.bs
```diff
@@ -844,16 +844,28 @@ The [=effective origin=] of an {{XRSpace}} can only be observed in the coordinat
...
+        <dd> Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |pose|'s {{XRPose/emulatedPosition}} to <code>true</code> if the relationship between |space| and |baseSpace| is only able to be evaluated via [=3DoF=] tracking but a computed offset was applied to simulate any amount of [=6DoF=] motion, or if the underlying tracking system otherwise indicates that the position was an emulated value.
+      <dt> Else if |space|'s pose relative to |baseSpace| has been determined in the past.
+        <dd> Set |transform|'s {{XRRigidTransform/position}} to the last known location of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
```
thetuvix: Nit: "position" rather than "location" here - we are otherwise consistent on using "position", as "location" may imply rotation as well.

index.bs
thetuvix: The [latest commit](0366de3403692e347a8a7e0ee0539d385eba9b8a) I see in this PR is from before my comment.  Did your latest changes end up getting pushed?

thetuvix:
The [latest commit](0366de3403692e347a8a7e0ee0539d385eba9b8a) I see in this PR is from before my comment.  Did your latest changes end up getting pushed?

toji:
Ugh. Sorry! I mixed up my remotes when pushing last time it seems. PR now has the latest changes.

index.bs
```diff
@@ -841,16 +841,35 @@ The [=effective origin=] of an {{XRSpace}} can only be observed in the coordinat
...
+    <dl class="switch">
+      <dt> If the tracking system provides a [=6DoF=] pose:
+        <dd> Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |transform|'s {{XRRigidTransform/position}} to the position of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |pose|'s {{XRPose/emulatedPosition}} to <code>false</code>.
+
+      <dt> Else if the tracking system provides a [=3DoF=] pose or indicates that the pose position is emulated:
+        <dd> Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |transform|'s {{XRRigidTransform/position}} to the tracking system's best estimate of the position of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=]. This MAY use a computed offset such as a neck or arm model. If a position estimate is not available, the last known position MUST be used.
```
thetuvix: "If a position estimate is not available", would we have entered this branch?

thetuvix: I see - this is for the first case, where the system provides only an orientation with no position.

thetuvix: There's a missing case here for a system that truly does just give "3DoF poses" for some space relation, never providing positions.  For example, if a 3DoF controller orientation is provided relative to the world orientation, but with no position at all.

Is it the job of the UA to then synthesize some position in that case, e.g. at the hip?  I guess it must be, since we have no way to express a position with no orientation?

thetuvix:
Thanks again for all the work to pull this together!  I have one comment above about a little piece that may be missing, but I'd be fine to just integrate this the way it is now.

To this point:
> I finally decided to try stripping away the non-normative bits and just see what it would look like if I only wrote up the actual functional changes and, well... it's actually pretty tiny. To the point where I'm concerned that I'm missing something. (Paging @thetuvix for a sanity check!) But then again the point of this was never that is was complex, just that we wanted to ensure developers were enabled to do the right thing in every situation.

That was indeed the conclusion I came to myself after all the analysis in https://github.com/immersive-web/webxr/issues/243#issuecomment-454282109: What we'd finally converged on is a set of fairly simple (though easy to design wrong!) set of base spatial tracking semantics for core WebXR that makes it easy for apps to assemble all the various VR and AR behaviors we've observed in practice.  Most of the non-normative text in the explainer then demonstrates how to assemble those various behaviors from the core building blocks, along with the tracking loss/recovery effects that naturally follow on from the spec language.  To me, that simple app assembly is preferable to an alternative where we needed lots of fiddly spec text to manage lots of stateful modes, since it means the logic sits in app code where it will implicitly behave the same way across UAs!

thetuvix:
So what's the next step here?  Should you or I merge these commits into my existing PR?  Or do you want to bring this in as a separate PR?

thetuvix:
I suppose I'd need to merge your commits, since I have rights to my PR branch.  Let me know if that's how you want to proceed and I can do that over the weekend.

toji:
If you want to rebase your branch on top of the latest master, then [this PR against your repo](https://github.com/thetuvix/webxr/pull/1) should be able to land cleanly on top of it. That would preserve the cleanest history of who worked on what.

thetuvix:
Thanks!  Unfortunately, I'd already been merging with upstream rather than rebasing within my PR, and so a rebase now was going to make things messier - I dove further down the merge rabbit hole and got your commit merged into https://github.com/immersive-web/webxr/pull/559.

I think we can close this editor-collar PR now.

Thank you so much for pulling this together!

PR #44, opened 2019-06-14
Roughing in the idea described in 689

NellWaliczek:
Ok, so clearly there's a ton of stuff to be added here (including all of the precise details john listed in his document), but I wanted to lay down a quick rough sketch of the combination of his work with josh's proposal.  As we've talked about before, the permissions requests happen at session creation and are optional because inline sessions won't always require them.  I'm ok with making them required and just having the permissions array be non-required or empty, though. 

This isn't intended to be a complete list of enumerated permissions, though from what i can tell from John's doc we probably will need a "spatial-tracking-unbounded" or some such as different from just "spatial-tracking" and will need to update the spatial-tracking-explainer.md accordingly.

Given that the AR stuff is still in the explainer, it might be worth adding an example "camera" since that's an easy one in the existing list of permissions requests?  Or should we not touch that with a 10ft pole?

When it comes to the actual spec, I'd like to have a section that very clearly states what's protected by each of the permissions in the list.  As well as a clear statement about whether we're going to allow for any non-permissions style mitigations for things that would otherwise require a permission?  A weird case is  IPD, though.  That one feels like we may need to call it out separately for Z-space for example if they allow it to be configurable?

Anyway, like I said.. super rough.


explainer.md
```diff
@@ -343,6 +375,17 @@ xrSession.addEventListener('end', onSessionEnd);
...
 If the UA needs to halt use of a session temporarily, the session should be suspended instead of ended. (See previous section.)
 
+## Privacy, security, and user consent
+
+### Privacy
+
+### Security
+
+### User consent
+For data that must be protect by user consent such as spatial tracking or non-quantized information that could lead to fingerprinting (ex: room bounds or inter-pupilary distance) developers must indicate their intent to use this data by passing the request into session creation.  User agents may choose to request user consent at session creation the developer-requested permissions, and if so the UA must not request permission again when the developer attempts to use features protected by the originally requested permissions.  Alternatively, user agents may choose to defer requesting user consent until developers attempt to use features that depend on the initially enumerated permissions.  If the request was deferred, the UA must instead request consent immediately prior to returning data protected by the permission.  Regardless of the timing of the consent request, if consent is granted, the features will work; if consent is not granted they will not.
```
toji: > If the request was deferred, the UA must instead request consent immediately prior to returning data protected by the permission.

Concerned about this wording, because it seems to imply that I could ask for a reference space that needed consent, but then not actually see a prompt until I did a `getPose()` with it.

NellWaliczek: oh, good call.  the intent was to prompt at requestReferenceSpace() if not done at session creation.  let me think about how to reword that.

explainer.md
```diff
@@ -137,7 +154,7 @@ function beginXRSession() {
 }
 ```
 
-In this sample, the `beginXRSession` function, which is assumed to be run by clicking the "Enter VR" button in the previous sample, requests an `XRSession` that operates in `immersive-vr` mode. The `requestSession` method returns a promise that resolves to an `XRSession` upon success. When requesting a session, the capabilities that the returned session must have, including it's XR mode, are passed in via an `XRSessionCreationOptions` dictionary.
```
toji: Oops.

explainer.md
```diff
@@ -397,15 +417,39 @@ function beginInlineXRSession() {
         // Inline sessions must have an appropriately constructed WebGL layer
         // set as the baseLayer prior to rendering. (This code assumes the WebGL
         // context has already been made XR compatible.)
-        let glLayer = new XRWebGLLayer(session, gl, { compositionDisabled: true });
```
toji: This was fixed in webxr/, so you'll pick it up when rebasing.

explainer.md
```diff
@@ -397,15 +417,39 @@ function beginInlineXRSession() {
...
+      })
+      .catch((reason) => { console.log("requestSession failed: " + reason); });
+}
+
+function onSessionStarted(session) {
+  session.requestReferenceSpace("local").then(onReferenceSpaceReady);
+}
+```
+
+The UA must not reject requests for an inline session unless the page's feature policy prevents it. `navigator.xr.supportsSession()` can still be used if a page wants to test if inline session are allowed.
```
toji: > The UA must not reject requests for an inline session unless the page's feature policy prevents it.... **or required permissions were rejected?**

Seems like we shouldn't imply that inline always succeeds regardless of permission requests.

privacy-security-explainer.md
```diff
@@ -0,0 +1,78 @@
...
+* **TODO** add more explanation of why this problem is harder on the web than for installed applications
+* **TODO** summarize the categories of threats in 1-2 paragraphs
+
+# Mitigations
+
+User agents must apply at least one mitigation to protect sensitive user data and interactions within an XR session. The appropriate mitigation will depend on the type of data being protected.
+* **TODO** Explain this a bit more
+
+## User initiated actions
+* **TODO** What is a user initiated action
```
toji: This should simply be a link off to the relevant bit of HTML spec, right?

privacy-security-explainer.md
```diff
@@ -0,0 +1,78 @@
...
+* **TODO** Explain why developers might want to check consent status (aka. you're about to see a pop up)
+* **TODO** If consent duration is set to be the length of a session, investigate using the Permissions API to enable this by adding a Permissions object to XRSession. It's `query()` API would behave similarly to the one on `navigator.permissions`, but instead checks for permissions granted for the duration of the session.
+
+### Permissions API
+* **TODO** Explain what the Permissions API is
+* **TODO** Explain the relationship between querying the consent status of XR permissions and the Permissions API
+
+### Non-XR permissions
+* **TODO** Maybe merge this with the section immediate above this?
+* **TODO** Explain that things like the microphone and geolocation which are unrelated to XR can cause prompts for user consent.
```
avadacatavra: Is microphone really unrelated to XR? Since text input in XR isn't very ergonomic, we've been pretty focused on getting voice input integrated. It seems more along the lines of camera/accelerometer/gyroscope, where it's not solely XR

NellWaliczek: yeah so this is a challenge I've been unsure how to think about... strictly speaking, microphone isn't an XR "feature".  It might be something used during an XR session, but isn't strictly an XR feature.  And I'm super aprehensive about the inputs into requestSession becoming a laundry list of all features that might ever request permissions.  
I attempted to strike a balance by the content of the next two bulleted items, but it does create a potential difference in how code should be written for browsers which support mid-session consent and those which don't.  Which raises an even more important question: "Is the implementation of a trusted UI a requirement to implement WebXR?"  I'll add that one to the Trusted UI section above.  But short of that, I'm not sure what other options we have design-wise.  

All that said, even if it was possible to request microphone access as part of session creation, it could still be denied by the user!  Which means sites will need to handle that progressive enhancement anyway, yes?

privacy-security-explainer.md
NellWaliczek: yep.  just hadn't dug it up yet and didn't want to forget.

explainer.md
NellWaliczek: Good call.

explainer.md
```diff
@@ -564,10 +612,46 @@ partial interface Navigator {
...
+
+dictionary XRReferenceSpaceFeatures {
+  XRFeatureConfiguration local;
+  XRFeatureConfiguration localFloor;
+  XRFeatureConfiguration boundedFloor;
+  XRFeatureConfiguration unbounded;
+}
+
+dictionary XRInputSourceFeatures {
+  XRFeatureConfiguration gamepads;
```
toji: 😒

I'm not eager to add barriers to features/modules like this unless we've found there's an explicit privacy issue. The one scenario that I could see this being valuable in, though, is if we want to expose gamepads to inline sessions. (But I don't see that being necessary)

spatial-tracking-explainer.md
```diff
@@ -266,10 +266,33 @@ Inline sessions, by definition, do not require a user gesture or user permission
...
+      {
+        referenceSpaceType: "local-floor",
+        requirement: "required"
+      },
+      {
+        referenceSpaceType: "bounded-floor",
+        requirement: "optional"
+      }
+    ]
+  }
```
toji: Nit: Older syntax.

privacy-security-explainer.md
```diff
@@ -0,0 +1,143 @@
...
+}
+
+dictionary XRPermissionDescriptor : PermissionDescriptor {
+  required XRSessionMode mode;
+  XRSessionOptions options;
+}
+
+// If we go this route
+partial interface XRSession {
+  readonly attribute Permissions permissions;
```
toji: Wouldn't be surprised if you've thought the same thing, but wanted to make sure to record my opinion on this: I absolutely believe that this aspect of the API (querying permission status) is necessary, and I absolutely believe we should be feeling out how it can work here. But given that it seems like one of the areas we're both the least certain about I'd also assert that this seems like additive API surface and isn't critical to the VR complete milestone in the same way that the `XRSessionOptions` are. As such unless we land on a solution for it we're completely comfortable with within the next couple of weeks I'd highly suggest deferring this part of the API (with the intent to have a solution by CR.)

explainer.md
```diff
@@ -137,7 +154,7 @@ function beginXRSession() {
 }
 ```
 
-In this sample, the `beginXRSession` function, which is assumed to be run by clicking the "Enter VR" button in the previous sample, requests an `XRSession` that operates in `immersive-vr` mode. The `requestSession` method returns a promise that resolves to an `XRSession` upon success. When requesting a session, the capabilities that the returned session must have, including it's XR mode, are passed in via an `XRSessionCreationOptions` dictionary.
+In this sample, the `beginXRSession` function, which is assumed to be run by clicking the "Enter VR" button in the previous sample, requests an `XRSession` that operates in `immersive-vr` mode. The `requestSession` method returns a promise that resolves to an `XRSession` upon success. In addition to the `XRSessionMode`, developers must supply an `XRSessionOptions` dictionary containing the capabilities that the returned session must have.
```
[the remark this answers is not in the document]
NellWaliczek: doh.

explainer.md
```diff
@@ -121,13 +121,30 @@ async function checkForXRSupport() {
...
 function beginXRSession() {
+
+  /* TODO
+      * Should we implicitly add "viewer", "local", and "local-floor" as "required" for "immersive-vr" so this line wouldn't be necessary?
+      * Should "immersive-ar" be the same as "immersive-vr"?
+      * Should "inline" have "viewer" default to "required" and nothing else?
+      * What would this mean for diorama reference spaces?
+  */ 
+  const sessionOptions = {
+    referenceSpaces: { "local-floor": { requirement: "required"} }
```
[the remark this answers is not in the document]
NellWaliczek: Oh boy I totally forgot to go back and up the sample code.  In a bunch of places haha

explainer.md
NellWaliczek: Yeah I added this one because i wanted to talk to you about it.  

1. I wasn't proposing it for *user consent* purposes; if we add it, it would be for feature requirement purposes.  For example, if you're building a painting application which requires more than a basic aim+select, it'd be unfortunate to create a whole immersive session just to tell users "ah sorry, no can do".  Specifically, this wouldn't gate on the *presence* of gamepad-based motion controllers, but rather the capacity to have them.  *shrugs* Not super attached to the idea, but wanted to make sure we'd thought it through.

2. As I was mulling over this, it occurred to me that we could have this (and hands) be marked as "optional" by default for immersive sessions and "not-requested" by default on inline.  But that also got me thinking... how many "if immersive than this default, but if inline that default" are going to be understandable to developers?  I mean we can fairly easily define two sets of defaults in spec text, but will developers know what to expect when they go to use it?

privacy-security-explainer.md
NellWaliczek: Yep, there's a reason I haven't been iterating on that part.  And from what I can tell, @avadacatavra has the same POV, yes?

PR #45, opened 2019-07-01
[WIP] Spec text for required and optional features

toji:
A bunch of handwaving here about how consent actually works, but the basic algorithms are in place. Uploaded here to serve as a discussion reference point if needed.


toji:
Now moved to webxr/749

